PR #1, opened 2018-12-04
angular 7 upgrade

whyboris:
👍 App upgraded to `angular 7`

_no behavior should change for the user_

PR #2, opened 2018-12-07
Thumbnail Improvements

cal2195:
👍 Ready to go!

This PR will add many improvements to the thumbnail system:

- [x] Change thumbnail id's to be hashes of the original files.
- [x] Use a single thumbnail sheet rather than 10 individual thumbnail files.
- [x] Allow resumable thumbnail generation. ( Depends on #5 )
- [x] Don't generate thumbnails if they already exist.
- [x] ~Add backwards compatibility to previous versions. (?)~ See #10.
- [x] ~Fix the demo content.~

Some notes about this PR:

- I've removed (hopefully) all references to the jpg number system.
- It currently generates 10 thumbnails as before, and then combines them - this could possible be improved to reduce the number of read and writes.
- I'm unsure whether a compatibility script is worth it, as it will require reading and writing all the thumbnails again anyway - it would be slightly quicker, but not sure by how much and adds complexity. If not, backing up the old files by renaming the folder and library to `.bak` would be a good way to go.
- As predicted, the single thumbnail sheet works perfectly on the thumbnail view, but has awful performance on the filmstrip view! 😢 It appears to be related to the animations, so I've removed them from the filmstrip view - not as pretty, but at least it performs well!
- I'm going to test the performance on a larger number of files when I'm home.

Any and all comments accepted! 😄


whyboris:
Fantastic! 

Thank you for your work -- I'll try to review everything by Sunday (max 3 days).

cal2195:
How does the demo content work/is it relevant? 😄

whyboris:
Demo content was meant to be an online version of the app -- as a demo of how users can interact with it. This example from February 2018: http://dev.yboris.com/vh/

I am thinking showing an online demo to users might cause more confusion, so I'm not sharing it publicly (and the demo is of the app version 1.0.0). Feel free to ignore these files without any updating.

Ps - to get the demo to work, just run `index.html` once you have the built app in the _dist_ folder 😉

main-support.ts
```diff
@@ -1,12 +1,26 @@
...
 
+/**
+ * Hash a given file using it's file name and size
+ * md5(file.name + file.size)
+ * @param fileName  -- file name to hash
+ * @param fileSize  -- file size to hash
+ */
+export function hashFile(fileName: string, fileSize: number): string {
+  // make the magic happen!
+  let hash = hasher('md5').update(fileName + fileSize.toString()).digest('hex');
```
whyboris:
```suggestion
  const hash = hasher('md5').update(fileName + fileSize.toString()).digest('hex');
```
Please consider using `const`

whyboris:
Somewhere in the process, the file size information stopped getting extracted (will look into your code to see if I can spot what happened).

main-support.ts
```diff
@@ -1,12 +1,26 @@
 import * as path from 'path';
+import mergeImg = require('merge-img');
```
whyboris: Unsure about imports, I think this is preferred:
```suggestion
import * as mergeImg from 'merge-img';
```

src/app/components/pipes/file-size.pipe.ts
```diff
@@ -6,12 +6,12 @@ import { Pipe, PipeTransform } from '@angular/core';
...
-   * @param numOfMb
+   * Return size of file formatted as XXX{mb,gb}
+   * @param sizeInBytes -- file size in bytes
    */
-  transform(numOfMb: number): string {
-    if (numOfMb) {
-      const rounded = Math.round(numOfMb);
+  transform(sizeInBytes: number): string {
+    if (sizeInBytes) {
+      const rounded = Math.round(sizeInBytes / 1000000.00);
```
whyboris: This makes the size appear as `0mb` for every video -- on old _vha_ files and on new ones created with this branch.
```suggestion
      const rounded = Math.round(sizeInBytes);
```

whyboris:
Compared to the _master_ and _angular7_ branch, for some reason the App uses a lot more RAM (500mb+ compared to ~130mb) on startup. 

While some people have requested automatic updating of screenshots on start, I would prefer it would not be the default, and only available (if you think it's worth it) with a setting in the menu.

In part it's a personal preference, but in part it could cause some trouble: making the app unresponsive/slow while a user just wants to use it. There are likely other scenarios where auto-update is not preferred.

Do you think it would be possible to have this behavior disabled until a user enables it in the settings?

whyboris:
While some people have requested automatic updating of screenshots on start, I would prefer it would not be the default, and only available (if you think it's worth it) with a setting in the menu.

In part it's a personal preference, but in part it could cause some trouble: making the app unresponsive/slow while a user just wants to use it. There are likely other scenarios where auto-update is not preferred.

Do you think it would be possible to have this behavior disabled until a user enables it in the settings?

whyboris:
I'm unsure if it's possible at all, but is there a way to bypass writing the 10 images to disk, and then reading the 10 images from disk to create the single image.

[`merge-img`](https://github.com/preco21/merge-img) interface seems to allow a *buffer* -- so perhaps we could keep the images in RAM and write once? That should make things faster (though unsure how much).

You said you've found a way to not use the `fluent-ffmpeg` which might mean we'll have a way to save the 10 screenshots to *buffer* rather than disk?

src/app/components/home/thumbnail/preview.component.ts
```diff
@@ -29,7 +29,7 @@ export class PreviewComponent implements OnInit, OnDestroy {
   @Input() title: string;
 
   hover: boolean;
-  currentlyShowing = 1;
+  currentlyShowing = 0;
```
whyboris: Do you prefer the first screenshot?

I've found that usually the screenshot from the 5% mark in the video often has useless content (in a cartoon it's often the opening credits - at worst rendering all screenshots from a cartoon series the same).

src/app/components/pipes/file-size.pipe.ts
cal2195: Since we're now hashing the file name and file size, I'd really like the actual file size to be in the library file, rather than the rounded version.

Especially since 1 calculation on a system that can do 3 billion a second isn't much overhead! ;) Reading in the thumbnail to display already dwarfs any effect this would have.

cal2195:
> I'm loving so much about this!
> 
> One minor note - the new screenshots are much larger than the sum of the previous files. In one randomly-chosen example the previous 10 screenshots weighed in at 214kb, while the new one was 1,150kb (I checked the file, the JPEG quality was set to 100). I suspect we just need to pass in a lower jpg compression setting somehow 👍
> 
> I looked at _merge-img_ but it doesn't seem to have any such options. It uses _jimp_ which seems to have something related to _quality_ but I don't know how to [access it](https://github.com/oliver-moran/jimp/tree/master/packages/type-jpeg).
> 
> Perhaps, we'll have to import `jimp` as a dependency too, and then before you do `img.write` we could use [something like this](https://github.com/oliver-moran/jimp/tree/master/packages/jimp#installation) to set the quality.

This should be fixed now with the new ffmpeg - I'm getting 61kb files here, can you confirm?

cal2195:
> Compared to the _master_ and _angular7_ branch, for some reason the App uses a lot more RAM (500mb+ compared to ~130mb) on startup.

Can you confirm if this is still an issue?

cal2195:
> While some people have requested automatic updating of screenshots on start, I would prefer it would not be the default, and only available (if you think it's worth it) with a setting in the menu.
> 
> In part it's a personal preference, but in part it could cause some trouble: making the app unresponsive/slow while a user just wants to use it. There are likely other scenarios where auto-update is not preferred.
> 
> Do you think it would be possible to have this behavior disabled until a user enables it in the settings?

I've opened #15 to track this! 👍

cal2195:
> I suspect it's possible to bypass writing the 10 images to disk, and then reading the 10 images from disk to create the single image.
> 
> [`merge-img`](https://github.com/preco21/merge-img) interface seems to allow a _buffer_ -- so perhaps we could keep the images in RAM and write once? That should make things faster (though unsure how much).
> 
> You said you've found a way to not use the `fluent-ffmpeg` which might mean we'll have a way to save the 10 screenshots to _buffer_ rather than disk?
> 
> What are your thoughts on this?

Haha, thanks for this! 👍 This prompted me to go and do some more research into ffmpeg, and I now have both image and video thumbnails being extracted with a single command into a single file! 🎉 

The image command has been added to this PR! Thanks again! 😄

main-support.ts
```diff
@@ -446,15 +481,12 @@ function extractMetadataForThisONEFile(
       const origHeight = metadata.streams[0].height;
       const sizeLabel = labelVideo(origWidth, origHeight);
       const width = Math.round(100 * origWidth / origHeight) || 169;
+      const fileSize = metadata.format.size;
```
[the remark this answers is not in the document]
cal2195: Same as above! 👍

src/app/components/home/thumbnail/preview.component.ts
cal2195: Ha, it feels a little funny to me not to start at the first image, but you have a good point! 👍

whyboris:
Amazing! I will have time after work on Friday (tomorrow) to try this out.

cal2195:
Don't merge yet, optimising thumbnails!

cal2195:
That last commit majorly improves thumbnail extraction speed! 👍 
I am noticing an issue with mkv files though, where the same first frame is used for all the thumbnails... more investigation is needed!

whyboris:
I wonder if _subtitles_ have something to do with the `mkv` problems you're having.

The 1min-long video extracted the first screenshot, just before any subtitles kicked in. I haven't confirmed, but I think all the _mkv_ files I tried this on (just a few in my case) - they all had subtitles!

maybe the `-sn` flag can help?

cal2195:
After some poking around with `ffprobe`, it turns out with `.mkv` files in particular, they don't output a duration in the streams collection! But they do in the format collection! 👍 And from the looks of it, all other files also have it in the format collection, so I've switched to using that by default!

The bug was caused because a duration of 0 meant that the step was also 0, so always picked the first frame!

cal2195:
Can you confirm this works for you too? 😄

whyboris:
Fantastic! _mkv_ works perfectly now!

Details: 43 videos imported in 1min 23sec! Average _jpg_ weight is 420kb (at 300px height per screenshot). 

App uses around 780mb of RAM instead of the usual ~150mb. I suspect it has to do with multiple loading of the images. I have ideas about getting that number down - can be done after this PR merges.

whyboris:
If you have the time, could you spot-check the _angular7_ #1 PR - once you approve it I'll merge it in, and then we can merge this one on top. I expect only about 2 lines of merge conflicts then 👌

cal2195:
In the middle of it right now! 😉

cal2195:
> App uses around 780mb of RAM instead of the usual ~150mb. I suspect it has to do with multiple loading of the images. I have ideas about getting that number down - can be done after this PR merges.

I'd assume this has to do with:
- on the thumbs view, the images are 10 times larger (each image loads all 10, rather than just loading when you hover over)
- on the film strip view, each strip is 10 images, each of which has all 10 screenshots - not sure if chrome reuses images here.

I'm sure there's some optimisation we could do, but merging first would be nice! 👍

whyboris:
It looks like when running the app in development mode the RAM usage is higher. I built the app and installed it with your branch, and it looks to be around ~165MB of RAM -- sorry for the false alarm 😓

cal2195:
Oh nice! 👍 Should be good to go then! 😄

whyboris:
I'm happy to merge this PR once the merge conflicts have been resolved 😁 

Once merged in, would you mind if I fix up some minor things (e.g. `var` -> `let`; `let` -> `const`; etc -- in a few places) - it's a bit easier than going back and forth with requests 😅

Issue #6, opened 2018-12-10
Prompt for root folder when saved root folder doesn't exist


wAuner:
If your videohub is on a SMB share the app saves the path in windows format. On mac and linux this leads to the problem, that if you click on a video in the app, the video will not open in your default video player.

The solution would be to save the filepaths in the .vha file in a different format, depending on the operating system of the client. On Unix systems this will be the mounting point (mac: /Volumes, Linux: /Media/username/...) 
Example:
For me this was "\\\\{name_of_network_share}\\Plex_videos\\video_folder” which works on windows.
I tried it on mac and changed that path manually to the correct mounting point on my system.
In this case “/Volumes/Plex_videos/video_folder”. Now when I click a video, it opens the video in VLC as intended.

cal2195:
Do you mean if you create a library on Windows, and then open it on Mac/Linux?

cal2195:
If so, then yes, this won't work - it's not an easy fix as you could mount the directory anywhere on your system, and even if you mount it in the same folder like Volumes, mac can change the name if there's other mounted shares.

One solution would be to detect if the video folder is missing, and allow you to browse to the correct mount point.

If we were to have multiple video paths for the root folder, it would be better to have it per system, rather than per OS, as the path isn't guaranteed to be the same on the same OS.

wAuner:
Good question. I can‘t remember on what os  I created the hub.
Is the mounting point really that unpredictable? Shouldn‘t it be always mounted in the same place unless you mount it manually via terminal or have a directory with the exact same name mounted?

cal2195:
Those should be the only two cases, but they're cases we need to take care of! Especially on linux, you're likely to mount it where ever you like, and each linux distro uses different mount points (/mnt, /media, /run) so we can't predict it.

And I've had MacOS mount my folder in folder-1 because the previous mount didn't unmount properly.

I'd rather let the user choose, rather than guess and get it wrong. Alternatively you could symlink all the folders to a common place! 😅

cal2195:
Also, this would fix a nice issue of moving your library too! 😄

cal2195:
The problem with that is that you don't have to keep the vha file with the actual video folder - you can choose to change it in the creation wizard!

I guess if you keep them the same in the creation wizard, we could make the assumption that files are relative to the vha file... Means you couldn't move them apart though...

cal2195:
[...]

PR #8, opened 2018-12-12
Allow for an input folder relative to the vha file location

### LINKED ISSUE #7: Use relative input folder when input folder and vha location are the same
When both the input folder and vha save location are the same, we should use relative directories to allow the library to be moved easily.

This will also allow the library to be opened on multiple systems when mounted at different locations.

cal2195:
This PR fixes #7, allowing for an input folder to be relative to the vha file location. This means you can move the whole library and vha file to a new location, and everything will still work! 😄

*Existing vha files will be updated when new files are added via a rescan.*

This is achieved when the input folder is set to an empty string in the vha file, which causes the input folder to be set to the same as the vha file parent folder.

When writing the vha file, if the input directory is the same as the vha file folder, the input directory will be set to an empty string!

In the future, it would be nice to add multiple input folders, and improve this logic if all input folders lie relative to the vha file! But for now, this just adds a little bit of functionality, and I assume most people leave the vha file with the input folder! 😄

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -76,6 +76,11 @@ export function countFoldersInFinalArray(imagesArray: ImageElement[]): number {
  * @param done          -- function to execute when done writing the file
  */
 export function writeVhaFileDangerously(finalObject: FinalObject, pathToTheFile: string, done): void {
+  // check for relative paths
+  if (finalObject.inputDir === path.parse(pathToTheFile).dir) {
+    finalObject.inputDir = "";
+  }
+
   const json = JSON.stringify(finalObject);
   // write the file
   fs.writeFile(pathToTheFile, json, 'utf8', done);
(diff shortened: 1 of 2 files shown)


whyboris:
Does this mean that if the user moves the `vha` file with the `vha-hub-name` folder the app will no longer find the folder it was tied to?

This isn't a problem -- I suspect that it may be worthwhile (later) implement a GUI feature to point to the correct location (if the folder is ever moved). The current _master_ has a similar problem -- moving the _folder_ where all the videos are also breaks the link. 🙆‍♂️

cal2195:
That is correct, as is the same in master! The only thing this changes is it allows you to move _everything_ if you keep the vha file with the videos, and it will still work - it only adds functionality!

I have #6 tracking your idea already! 😄

cal2195:
If you already store the vha file separate from your videos, nothing changes! 👍

cal2195:
We've come a long way! 😄

PR #20, opened 2018-12-16
Translation system for any language

whyboris:
Adds a translation system for the entire app.

How to add another language:
1. just make a copy of the `en.ts` file, rename (e.g. `jp.ts`) 
2. translate all strings
3. import file into `home.component.ts`
4. add to `changeLanguage` method

🚀 and you're done!


whyboris:
There are some lingering strings that need translation. Will handle later this week.

README.md
```diff
@@ -46,8 +46,10 @@ This software would not be possible without the tremendous work by other people:
...
  - [Angular](https://github.com/angular/angular)
  - [Electron](https://github.com/electron/electron)
+ - [FFmpeg](https://www.ffmpeg.org/)
  - [angular-electron](https://github.com/maximegris/angular-electron)
  - [angular2-virtual-scroll](https://www.npmjs.com/package/angular2-virtual-scroll)
- - [fluent-ffmpeg](https://github.com/fluent-ffmpeg/node-fluent-ffmpeg)
  - [@ffmpeg-installer/ffmpeg](https://www.npmjs.com/package/@ffmpeg-installer/ffmpeg)
  - [@ffprobe-installer/ffprobe](https://www.npmjs.com/package/@ffprobe-installer/ffprobe)
+ - [ngx-translate](https://github.com/ngx-translate/core)
+ - [fluent-ffmpeg](https://github.com/fluent-ffmpeg/node-fluent-ffmpeg) _(no longer used)_
```
cal2195: This has reminded me - we still use this for the ffprobe for metadata if I recall correctly... might be nice to change that and remove this dependency!

src/app/components/home/home.component.html
```diff
@@ -302,10 +303,22 @@ <h2>Change the app zoom level</h2>
       </button>
     </div>
 
+
+    <div class="horizontalDivider"></div>
+
+    <h2>{{ 'SETTINGS.changeLanguage' | translate }}</h2>
+
+    <div style="margin: 20px">
+      <select (change)="changeLanguage($event.target.value)">
```
cal2195: It might be here, but I didn't notice it:
Does this save and/or detect the language the user would like to use? 😄

README.md
whyboris: Good catch -- I didn't realize we still have _fluent-ffmpeg_ as a dependency. I don't mind keeping it, though perhaps if we use _ffprobe_ directly we could query just the relevant features of the file we want (resolution & length) rather than grabbing all metadata about it 👍

src/app/components/home/home.component.html
whyboris: It looks like we can do something like this:
`translate.use(browserLang.match(/en|ru/) ? browserLang : 'en');`
But I'm unsure how _Electron_ handles browser language.
It might be worth adding a modal if the app detects it's the first start that has a language selection dropdown.

I'll likely add that in a different PR (I prefer smaller PRs generally: since I'm not building to _production_ every time a PR lands, I can merge in smaller features).

cal2195: Created #26 to track this!

src/app/components/common/settings-object.interface.ts
```diff
@@ -1,5 +1,8 @@
+export type PossibleLanguage = 'en' | 'ru';
```
cal2195: Would this be better named `SupportedLanguage`?

src/app/components/home/home.component.html
```diff
@@ -254,12 +254,13 @@ <h2>{{ settingsCategories[i] }}</h2>
             *ngIf="settingsButtons[buttonKey].iconName !== 'icon-checkmark'"
             (click)="toggleHideButton(buttonKey)"
             class="eyeIconContainer icon {{ settingsButtons[buttonKey].hidden ? 'eye-closed' : 'eye-open' }}"
+            title="show / hide icon in the settings bar"
```
cal2195: Should this be translated? 😄

src/app/components/home/home.component.html
```diff
@@ -766,20 +780,22 @@ <h2>Import new hub</h2>
       [disabled]="futureHubName === ''"
       title="Please select the folder with your video files"
```
cal2195: Should this be translated? 😄

src/app/components/home/home.component.html
```diff
@@ -794,11 +810,12 @@ <h2>Import new hub</h2>
       [disabled]="wizard.selectedSourceFolder === '' || futureHubName === ''"
       title="Please select the folder where your Video Hub will be stored"
```
cal2195: Translated? (Maybe I'm missing something! 😆 )

src/app/components/home/tags/tags.component.ts
```diff
@@ -97,7 +97,7 @@ export class TagsComponent implements OnDestroy {
       this.currentAdding = '';
 
     } else {
-      this.showStatusMessage('tag already exists')
+      this.showStatusMessage('tag already exists');
```
cal2195: Translate? 😄

src/app/components/home/home.component.html
whyboris: Great catch!

src/app/components/home/home.component.html
whyboris: Missed this one too 😓

src/app/components/home/home.component.html
whyboris: Hooray! 😅

src/app/components/home/tags/tags.component.ts
whyboris: Another one 🔎

cal2195:
Looks great! 👍

PR #30, opened 2018-12-19
Replace fluent-ffmpeg with ffprobe exec command!

### LINKED ISSUE #25: Remove fluent-ffmpeg dependency
As we're only using fluent-ffmpeg now for ffprobe, we should consider removing it and calling ffprobe directly and just parsing the json results! :)

cal2195:
Removed `fluent-ffmpeg` dependency, and replaced it with our own `exec` of `ffprobe`! 👍 

Fixes #25!

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@ffmpeg-installer/ffmpeg": "1.0.17",
     "@ffprobe-installer/ffprobe": "1.0.9",
     "@ngx-translate/core": "11.0.1",
-    "angular2-virtual-scroll": "0.3.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -35,7 +35,6 @@
     "@ffprobe-installer/ffprobe": "1.0.9",
     "@ngx-translate/core": "11.0.1",
     "angular2-virtual-scroll": "0.3.1",
```
whyboris: please remove the `,` from the end of this line - 👍 else I get "_Failed to parse json_" on `npm start`

cal2195: Ah! From merging! Thanks! 👍

PR #32, opened 2018-12-19
Implement prompt for root folder when folder doesn’t exist!

### LINKED ISSUE #6: Prompt for root folder when saved root folder doesn't exist

cal2195:
All the PRs for @whyboris! 😅 

This PR prompts users to select the new location of the video root folder if the saved one doesn't exist! ✅ 

The only case that isn't handled now is if you move a relative `vha` file out of the video folder, as the current folder will always exist (the one the `vha` file is in!) 👎 

Fixes #6!

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -247,8 +247,8 @@ function openThisDamnFile(pathToVhaFile: string) {
   fs.readFile(pathToVhaFile, (err, data) => {
     if (err) {
       dialog.showMessageBox({
-        message: 'No such file found \n' +
-          pathToVhaFile,
+        message: 'No such file found:',
+        detail: pathToVhaFile,
         buttons: ['OK']
       });
       globals.angularApp.sender.send('pleaseOpenWizard');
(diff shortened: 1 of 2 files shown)


cal2195:
Interesting - I didn't think of that use case - what purpose would that serve? 🤔 

To support this use case, I can just add a "Continue anyway" button, which would continue with the current behaviour? ✅

cal2195:
Added such a button! 👍

whyboris:
Confirmed works beautifully on Windows and Mac 🥇

PR #33, opened 2018-12-21
Upgrade electron to v4.0.0 and implement Dirent objects for file listing!

### LINKED ISSUE #3: Use fs.Dirent objects, and extract metadata in the background
Currently, the app freeze while performing the first phase of the import (listing all the files recursively).

A large portion of this delay in phase 1 is due to `stat`ing all the files to see if they are directories. Alternatively using `fs.Dirent` objects, the following code would be much better:
```
// Recursively walk through a directory compiling ImageElements
  const walkSync = (dir, filelist) => {
    const files = fs.readdirSync(dir, {encoding: 'utf8', withFileTypes: true});

    files.forEach(function (file) {
      if (file.name) {
        if (!fileSystemReserved(file.name)) {
          // if the item is a _DIRECTORY_
          if (file.isDirectory()) {
            filelist = walkSync(path.join(dir, file.name), filelist);
          } else {
            const extension = file.name.split('.').pop();
            if (acceptableFiles.includes(extension.toLowerCase())) {
              // before adding, remove the redundant prefix: sourceFolderPath
              const partialPath = dir.replace(sourceFolderPath, '');
              // fil finalArray with 3 correct and 5 dummy pieces of data
              finalArray[elementIndex] = [partialPath, file.name, cleanUpFileName(file.name), '', 0, '', 0, 0];
              elementIndex++;
            }
          }
        }
      }
    });
```
See https://nodejs.org/api/fs.html#fs_fs_readdirsync_path_options

This will be a drastic improvement in speed, especially over networked file systems. Unfortunately, this is only added in node v10.10.0, which will need to be updated first.

Additionally, it would be much better if the second phase (metadata extraction) was performed in the background like phase 3 (thumbnail generation). This is especially important with the addition of file hashes, which will require reading every file.

cal2195:
***Poof!*** 🎉 
This PR upgrades electron to v4.0.0, and uses Dirent objects for file listing which should _majorly_ speed up the file listing process! 👍 

Tested and works for me! 😄 

Fixes #3!

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -138,20 +138,20 @@ export function getVideoPathsAndNames(sourceFolderPath: string): ImageElement[] 
 
   // Recursively walk through a directory compiling ImageElements
   const walkSync = (dir, filelist) => {
-    const files = fs.readdirSync(dir);
+    const files = fs.readdirSync(dir, {encoding: 'utf8', withFileTypes: true});
 
     files.forEach(function (file) {
-      if (!fileSystemReserved(file)) {
+      if (!fileSystemReserved(file.name)) {
         // if the item is a _DIRECTORY_
-        if (fs.statSync(path.join(dir, file)).isDirectory()) {
-          filelist = walkSync(path.join(dir, file), filelist);
+        if (file.isDirectory()) {
+          filelist = walkSync(path.join(dir, file.name), filelist);
         } else {
-          const extension = file.split('.').pop();
+          const extension = file.name.split('.').pop();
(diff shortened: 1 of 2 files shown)


cal2195:
Otherwise, electron 3 will just still return strings from the `readdirSync` call, and string.name won't exist!

whyboris:
whoops -- I realized I forgot to `npm install` 😓 sorry

whyboris:
I keep forgetting that it's _electron_ that will run node (which is its own version), not what I currently have on my machine 😅

PR #40, opened 2018-12-26
Folder-only view

whyboris:
This PR adds a new view to the app: *Folder-only view*

It will look like the file view, but will only show folders.
Clicking on the folder name should open the _file_ view showing all the files in that folder.


whyboris:
I'll think about how to add a `go up` and/or `go back` button 👍

Issue #38, opened 2018-12-26
Remember last window size after app closed


whyboris:
*Problem:*
Currently app opens to full screen every time. 

*Solution:*
Have the app remember the last window size.
Optionally add a setting to the menu.

whyboris:
This will be closed with this PR: https://github.com/whyboris/Video-Hub-App/pull/41

whyboris:
Should be fixed since PR #41 got merged ✅

PR #41, opened 2018-12-26
Remember window size and position

whyboris:
Restore window size and position for the app.

As a bonus, during loading, the app shows a small rectangle with an animation (like Photoshop and other software) and then opens to the user's last window size and position when ready.

If the screen resolution changes (and the app's top-left corner would otherwise appear off-screen according to previously-saved location), the app should reset location to (0, 0).


whyboris:
This PR will close: https://github.com/whyboris/Video-Hub-App/issues/38

whyboris:
I'll investigate flickering a bit later 👌 - I'll be doing finishing polish after building the app and testing it out then 😄

cal2195:
I noticed the flickering when it resizes, but I was using the dev version! I was just wondering as it might be a little less jarring if it started in the correct position/size, but it's not very important! 😄

PR #43, opened 2018-12-27
Better organized settings menu

whyboris:
First pass at it 🙆‍♂️ 

I also added the [keyboard shortcuts](https://videohubapp.com/faq.html)!

We will need better organization / labeling later, but this makes things much better already.


whyboris:
For an easier review, just add `?w=1` to the URL like this:
https://github.com/whyboris/Video-Hub-App/pull/43/files?w=1

This hides all the white-space changes (easier to see what's added / removed - most of the changes GitHub shows are just indentation because I've added wrapping elements around the buttons)

src/app/components/home/home.component.html
```diff
@@ -335,6 +336,33 @@ <h2>{{ 'SETTINGS.changeLanguage' | translate }}</h2>
 
     </div>
 
+    <!-- TODO -- TRANSLATE !!! -->
+    <div *ngIf="settingToShow == 4" class="shortcutMenu">
+
+      <span>Hold <em class="key modifierKey">Ctrl</em>
+        <ng-container *ngIf="macVersion">
```
cal2195: This doesn't work on the dev version - I assume it works if you build the app?

whyboris: Thank you for checking -- I'll double check this later. What I had expected was that if you change the `macVersion` to `true` (line 103 in `home.component.ts`) you should be able to see this line of text. The boolean is one of the manual changes needed when building the app for Mac (so that you get the native-looking close, maximize, minimize buttons) ✅

cal2195: Oh! I hadn't noticed that - maybe there's a way to detect the OS you're running on? Save changing it when building/testing the app! 💯

Issue #39, opened 2018-12-26
Auto play 10 screenshots on hover option


whyboris:
*Problem:*
Some users prefer to have the thumbnails view cycle through the screenshots without having to move the mouse

*Solution:*
Add a settings toggle that will allow the thumbnails to cycle without having to move the mouse.
If I remember correctly, the code for the cycling is already there, but disabled.

whyboris:
May be best to wait for the new filmstrip component branch to settle: 
https://github.com/whyboris/Video-Hub-App/tree/new-filmstrip
The `new-filmstrip` should make it easier to handle more than 10 screenshots once we start allowing users to extract more 👍

cal2195:
Would also be nice to add the `autoplay` attribute to the video previews in this case! 💯

cal2195:
(and the muted tag! 😂 )

whyboris:
The `new-filmstrip` PR is ready.
We could start work on this auto-play issue once the PR lands: https://github.com/whyboris/Video-Hub-App/pull/44

whyboris:
I think we should delay work on this feature until we've implemented the feature / option to scan more than 10 screenshots 👍

cal2195:
#76 has landed! :tada: Good to go!

whyboris:
Currently working on "full preview view" for the video - I can get this done right after 👍

whyboris:
I'll look into this now

whyboris:
Will look at it tomorrow 👍 
Will also see if I can clean up some old buttons (miscellaneous functionality) that are unsorted and possibly not working 😅

whyboris:
On my `to do next` list 🌊 
Disregard my past self from Feb 25 🤣 perhaps "tomorrow" meant something different in the past 😅

PR #44, opened 2018-12-27
New filmstrip

whyboris:
⚠️ PR is WIP ⚠️ 

Thank you @cal2195 for the `ffmpeg` extraction code to get the images arranged horizontally and resized properly with `hstack` 👍 

I'd love to add black bars between the screens, I suppose 3px would be the best size, do you have any preferences?

I'll try to code it up but might turn to you for help again:
https://video.stackexchange.com/questions/18419/add-black-or-other-color-padding-when-doing-vstack-or-hstack

Cheers!


whyboris:
Thank you for noticing the `;` - I typed out the changes you've made (to show me what to do) on your copy of the branch into my branch to get a better sense of what was happening - but should have been more careful 😅 - noticed it didn't work after the commit and fixed it 👍

whyboris:
I think it would make the math easier with fewer bugs in the app if we extracted all images onto a standard black background with consistent size regardless of the source video aspect ratio. 

I would opt for a `16:9` (~1.777) standard across all screenshots regardless of the underlying resolution. We would then only need to store a single number across the whole `.vha` file rather than individual widths for each video!

This would require a bit of tricky `ffmpeg` scripting, but seems like it's doable. We'll start with a black base background. For example, a user selects a height of 100px for all screenshots. We compute the desired output to be 100 * 1.777 = 178 and add 1px for a minimum border on each side (180px total). When extracting screenshots we'll match the height to 100px and keep the video's aspect ratio to output the video to the center of a 180px region (cropping anything that is over 178px wide).

The height (100px) and width (180px) are saved in the .vha file and every thumbnail and filmstrip preview simply uses those dimensions for computing which of the many screen-captures to show.

whyboris:
The upshot of this approach is that within the app we can simply show the relevant region of the horizontal strip (which will have the black bars around each image within the jpg).

whyboris:
I got a script to work, but it works by manually adding vertical stripes (no fun) ... would prefer to do it another way: pasting into the center a (possibly cropped) video screenshot, and attaching the output as a horizontal stack. But for now:
```
-ss 811.3636363636364 -i 'E:\Movies\someMovie.mp4' \
-ss 1622.7272727272727 -i 'E:\Movies\someMovie.mp4' \
-ss 2434.090909090909 -i 'E:\Movies\someMovie.mp4' \
-ss 3245.4545454545455 -i 'E:\Movies\someMovie.mp4' \
-frames 1 -filter_complex \
"[0:v]scale=300:-2[0]; \
color=black:5x162[0c]; \
[1:v]scale=300:-2[1]; \
color=black:5x162[1c]; \
[2:v]scale=300:-2[2]; \
color=black:5x162[2c]; \
[3:v]scale=300:-2[3]; \
[0][0c][1][1c][2][2c][3]hstack=inputs=7" \
'C:\temp\output.jpg'
```

whyboris:
There's a very similar question with a great answer on _StackOverflow_:
https://superuser.com/questions/547296/resizing-videos-with-ffmpeg-avconv-to-fit-into-static-sized-player

whyboris:
Branch now works as desired 🎉 
I'll now remove previous mentions and needless computations of screenshot widths 👍 
Currently there may be rounding errors (1px per scroll) when extracting screenshots not at 300px. Will fix.

whyboris:
Once this PR lands, this issue will be ready to work on: https://github.com/whyboris/Video-Hub-App/issues/39

whyboris:
⚠️ Minor bug -- current branch doesn't work if the `.vha` file name has any spaces in it 😅 
Will fix 👌

main-support.ts
```diff
@@ -243,27 +243,37 @@ export function takeTenScreenshots(
...
+  let allFramesFiltered = '';
+  let outputFrames = '';
+
+  // Hardcode ~16:9 ratio
+  const ssWidth: number = Math.ceil(screenshotHeight * 16 / 9);
+  const ratioString: string = ssWidth + ':' + screenshotHeight;
+
+  // sweet thanks to StackExchange!
+  // https://superuser.com/questions/547296/resizing-videos-with-ffmpeg-avconv-to-fit-into-static-sized-player
+  const fancyScaleFilter = 'scale=' + ratioString + ':force_original_aspect_ratio=decrease,pad=' + ratioString + ':(ow-iw)/2:(oh-ih)/2';
```
cal2195: I'm getting an issue when I try to use a thumbnail size of 50px:
```
grep stderr: Stream mapping:
  Stream #0:0 (h264) -> scale
  Stream #1:0 (h264) -> scale
  Stream #2:0 (h264) -> scale
  Stream #3:0 (h264) -> scale
  Stream #4:0 (h264) -> scale
  Stream #5:0 (h264) -> scale
  Stream #6:0 (h264) -> scale
  Stream #7:0 (h264) -> scale
  Stream #8:0 (h264) -> scale
  Stream #9:0 (h264) -> scale
  hstack -> Stream #0:0 (mjpeg)
Press [q] to stop, [?] for help

grep stderr: [swscaler @ 0x7fd12ada9a00] deprecated pixel format used, make sure you did set range correctly

grep stderr: [Parsed_pad_1 @ 0x7fd129099800] Input area 0:0:89:50 not within the padded area 0:0:88:50 or zero-sized
[Parsed_pad_1 @ 0x7fd129099800] Failed to configure input pad on Parsed_pad_1

grep stderr: Error reinitializing filters!
Failed to inject frame into filter network: Invalid argument

grep stderr: Error while processing the decoded data for stream #9:0

grep stderr: Conversion failed!
```

main-support.ts
```diff
@@ -566,12 +576,11 @@ function extractMetadataForThisONEFile(
       const origWidth = metadata.streams[0].width;
       const origHeight = metadata.streams[0].height;
       const sizeLabel = labelVideo(origWidth, origHeight);
-      const width = Math.round(100 * origWidth / origHeight) || 169;
       const fileSize = metadata.format.size;
       imageElement[3] = hashFile(imageElement[1], fileSize);
       imageElement[4] = duration;  // 4th item is duration
       imageElement[5] = sizeLabel; // 5th item is the label, e.g. 'HD'
-      imageElement[6] = width;     // 6th item is width of screenshot in px (e.g. 150);
+      imageElement[6] = 0;         // 6th item WILL BE DEPRECATED AND REMOVED !!!
```
cal2195: We should leave this until we convert the array to objects! 👍

src/app/components/home/thumbnail/preview.component.html
```diff
@@ -7,44 +7,32 @@
...
-      class="hoverable"
-    >
-    </div>
+  <div
+    #filmstripHolder
+    class="video-box"
+    (mousemove)="mouseIsMoving($event)"
+    [ngStyle]="{
+        height: imgHeight + 'px',
+        'background-image': 'url(' + (hover ? fullFilePath : firstFilePath) + ')',
```
cal2195: I like this, apart from 2 things:

1. Can we make it so that the image doesn't flick back to the first image when you stop hovering?

2. We should sync up the image and video thumbnails (at least the first one) if we're going to do this!

cal2195:
> Branch now works as desired 🎉
> I'll now remove previous mentions and needless computations of screenshot widths 👍
> Currently there may be rounding errors (1px per scroll) when extracting screenshots not at 300px. Will fix.

This is happening for me even at 300px! 😕

main-support.ts
cal2195:
```
Input area 0:0:89:50 not within the padded area 0:0:88:50
```

Looks like rounding error!

main-support.ts
```diff
@@ -323,8 +333,8 @@ export function takeTenClips(
     concat += '[' + (current - 1) + ':v]' + '[' + (current - 1) + ':a]';
     current++;
   }
-  concat += 'concat=n=' + (totalCount - 1) + ':v=1:a=1';
-  args.push('-filter_complex', concat, saveLocation + '/' + fileHash + '.mp4');
+  concat += 'concat=n=' + (totalCount - 1) + ':v=1:a=1[v][a];[v]scale=' + screenshotHeight + ':-2[v2]';
```
cal2195: I just realised, this is now backwards (generates 300px wide, not high)!

Should be `-2:screenshotHeight`!

cal2195: I'll also confirm later whether we need -2 or -1!

main-support.ts
whyboris: Yes -- I'll just keep it at 0 as it's not needed. Will only remove once we convert to array of objects.

src/app/components/home/thumbnail/preview.component.html
whyboris: 1. I made the preview flick back to the original image on purpose as requested by a customer. But since the _settings_ menu is cleaner I can add more option toggles -- this one could be it.
2. I rather like that the video thumbnail image (the `-first.jpg`) is different from any of the ones in the filmstrip. I'm happy to keep them different.

cal2195: 1. Yes please! 👍 I like it staying where it is!

2. Hmm, it conflicts with option 1 though, if it doesn't reappear... maybe it isn't an issue.

whyboris: I'll put in a settings button that will allow users to choose between retaining the last-viewed thumbnail or to always come back to the first preview image.

What other fixes would you like to see before we can merge this in? (I know there's more polish and bug-fixing to do before this PR is a fully-operational feature, but I'd rather patch them with subsequent PRs rather than spend many more days on this one) 😃

cal2195: If your happy with bug fixes and changes later - merge away! 🎉 👍

whyboris:
Excellent! Thank you for the approval -- I'll merge this in now and get started on bug-fixing / improvement PR.

🎉

main-support.ts
whyboris: I copied this change from your other branch without looking into what it does -- feel free to edit anything in `takeTenClips` method as you see fit 👍

Issue #29, opened 2018-12-19
Library .vha structure and format


cal2195:
The current `.vha` file structure is a json of the following:
```
export interface FinalObject {
  hubName: string;      // the name of the hub -- for recently-opened
  ssSize: number;       // screen shot size -- so when you reimport it remembers your preference
  numOfFolders: number; // number of folders
  inputDir: string;     // later may support array of many input directories
  addTags?: string[];    // tags to add
  removeTags?: string[]; // tags to remove
  images: ImageElement[]; // see below
}

export type ImageElement = [string, string, string, string, number, string, number, number];
/*                            0       1       2       3       4       5       6       7
-------------------------------------------------------------------------------------------------------
index   type      description                   more info
-----   ------    -------------------------------------------------------------------------------------
  0     string    partial path to file,    <--- for opening the file, just prepend the `inputDir`
  1     string    full original file name, <--- for opening the file
  2     string    file name cleaned
                  of dots, underscores,
                  and file extension       <--- for searching
  3     string    hash of the file         <--- used for detecting changed files and as a screenshot
                                                identifier
  4     number    duration                 <-- duration of film as number
  5     string    depicting size           <-- e.g. `720`, `1080`, `SD`, `HD`
  6     number    width of screenshot      <-- e.g 124 etc
  7     number    fileSize in megabytes    <-- rounded to nearest MB

*/
```

Why are ImageElements arrays, rather than objects, so we could do `ImageElement.duration`?

If it's to save space in the saved file, we could compress the file (which would work very well due to the duplicated values across files eg. object names, input folder, resolution, width) at the cost of slightly less easy manual editing of the file?

Alternatively, using constants like `PATH_TO_FILE=0`, `FILENAME=1` so we can easily index the array would be nice!

Let me know if I'm misunderstanding something! 😄

whyboris:
I started out the structure out of convenience; resisted making the elements into objects (rather than arrays) because of redundant information that would be stored in the .vha file. Compression would solve that of course. If we use compression, I'd prefer the simplest _zip_ compression so users could unzip, edit, and re-zip if needed.

I like the constants idea - makes the code easier to read than `foo[4]`, but I'm also happy to change the structure to an array of objects. I'm also fine with keeping things as is -- what do you see as the biggest benefit(s) of changing?

The main down side would be no backwards compatibility unless we write a conversion script.

cal2195:
[...]

PR #47, opened 2018-12-28
Convert vha files to zip files!

cal2195:
This is a precursor to #29 - it now saves vha files as zip files, and loads them in correctly!

As a bonus, we could also store other data in this zip file! 💡 

I haven't looked into optimal compression settings yet, but that can be improved! 👍

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -2,6 +2,7 @@ import * as path from 'path';
 
 const fs = require('fs');
 const hasher = require('crypto').createHash;
+const JSZip = require('jszip');
 
 import { FinalObject, ImageElement } from './src/app/components/common/final-object.interface';
 import { acceptableFiles } from './main-filenames';
@@ -82,6 +83,7 @@ export function countFoldersInFinalArray(imagesArray: ImageElement[]): number {
   return finalArrayFolderMap.size;
 }
 
+
 /**
  * Write the final object into `vha` file
  * @param finalObject   -- finalObject
@@ -95,9 +97,18 @@ export function writeVhaFileDangerously(finalObject: FinalObject, pathToTheFile:
   }
 
   const json = JSON.stringify(finalObject);
+
   // write the file
-  fs.writeFile(pathToTheFile, json, 'utf8', done);
-  // CATCH ERRORS !?!!?!!
+  const zip = new JSZip();
+  zip.file('vha.json', json);
+  zip.generateNodeStream({type: 'nodebuffer', streamFiles: true})
+     .pipe(fs.createWriteStream(pathToTheFile))
+     .on('finish', function () {
+          // JSZip generates a readable stream with a "end" event,
+          // but is piped here in a writable stream which emits a "finish" event.
+          console.log(pathToTheFile + ' written!');
+          done();
+     });
 }
 
 /**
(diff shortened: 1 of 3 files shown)


whyboris:
Looks clean - I'll try it out within this hour and merge if it runs without issue 👍

whyboris:
Of course this branch would be a super-breaking change for old files. That's not too much of a problem since users will have to re-import their screenshots anyway. But with this change, at the moment, users won't be able to pen any of their old files either - an error would occur.

At the moment I get this error: 
>UnhandledPromiseRejectionWarning: Error: Can't find end of central directory : is this a zip file ? If it is, see https://stuk.github.io/jszip/documentation/howto/read_zip.html
> Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
Sadly it breaks the app in a way that it doesn't even go to the _Wizard_ menu to start a new hub.

whyboris:
It looks like it's possible to catch the error and re-use the old method of parsing the file ... but it may add more complexity. It's maybe OK to just declare 0 backwards compatibility with older `vha` files going forward (and I could have a tiny write-up for what people can do to manually move the _auto-tags_ if they have any).

I'm also wondering if without the compression the .vha files will be _that_ large -- you said you have tried a video directory with ~10,000 files. I'm guessing the `.vha` is under 10mb. 

So we _could_ just not compress the file in the end - saving some coding complexity and not requiring another dependency 😅

cal2195:
I just had a look - my uncompressed vha file is currently 1.7mb.

Rough calculations lead me to think that converting to objects would add about 2mb... (based on 200 characters added per file).

Do you think this is worth it?

whyboris:
If a person is using this app, they're already committing to hundreds of megabytes taken up by screenshots, I don't think anyone would mind a 10mb _vha_ file 😅 

I'm certainly not opposed to zip compression -- it just looks like it might not be a super-quick change (catching errors, handling old files, and testing to make sure there are no edge cases we missed) -- only work on this if you're really eager about this feature, not if you think it's instrumental to converting .vha from an array of arrays to an array of objects 😅

cal2195:
You have a good point - I'll leave it here for now, but we mightn't use it! 👍

cal2195:
Update:

A sample .vha file with lots of files is `6mb` uncompressed with the new format.

Compressed, this file is `1.3mb` (80% reduction).

cal2195:
@whyboris What are your current thoughts on zip compression?

whyboris:
I'm not opposed to it if it's something you'd like to implement 👍 
Do let me know if you would like any assistance other than a PR review 👍

whyboris:
Now that we're using `vha2` there's no conflict in having a completely different format than before (zipped _JSON_ rather than uncompressed _JSON_, that is). 

I still feel like compressing is added code complexity -- and saving hard drive space isn't important when the difference is 2MB vs 6MB. 

I'm thinking we either implement it before `2.0.0` or this feature would have to wait for `3.0.0` / if ever.

I'm leaning towards keeping things as is. @cal2195 -- please feel free to close the PR or let's discuss here whether we should get this feature in 👍 

In theory, there should only be 2 places where the new _zip_ code would be operating - when opening and when saving a file so it shouldn't be _too_ problematic ... but I rather like the hackability of the current file type (no need to unzip -- just open in notepad and hack away) 🤔

whyboris:
I'm closing this as I'm cleaning up for version 2 release.

@cal2195 -- please comment if you'd like me to reopen (or reopen it yourself if you have the repository permission 👍 )

PR #48, opened 2018-12-28
Filmstrip bugfixes

whyboris:
This PR is for minor cleanup and two small features:
- better thumbnail previews (_preview component_) - should have correct offset
- now there's a settings option whether the thumbnail screenshot should remain when mouse leaves, or the thumbnail should return to the first screenshot 👍 

There are some lingering issues since the last PR (#44) - but I'm tracking them with this issue: #50 

Please feel free to leave comments about other bugs in the current `master` in #50 👍 

This PR introduces breaking changes -- old `.vha` files will error out :trollface:

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -222,7 +222,7 @@ const exec = require('child_process').exec;
 /**
  * Take 10 screenshots of a particular file
  * at particular file size
- * save as particular fileNumber
+ * save as particular fileHash
  * @param pathToVideo  -- full path to the video file
  * @param fileHash     -- hash of the video file
  * @param screenshotHeight   -- height of screenshot in pixels (defaul is 100)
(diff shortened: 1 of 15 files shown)


cal2195:
Just an odd thing I've noticed - not sure what's caused it, but lately there's a 1 px line all the way along the top!

screenshot 2018-12-28 at 19 46 22

whyboris:
I've not looked into _dark mode_ in a while. Thank you for catching that -- I'll fix 👍

whyboris:
I just noticed, unsure how long this has been going on:

> [7592:1228/153458.305:ERROR:CONSOLE(5566)] "SyntaxError: Unexpected token m in JSON at position 614129", source: chrome-devtools://devtools/bundled/shell.js (5566)

Have you seen this before?

whyboris:
I fixed up the _thumbnail_ view to use percent offset rather than pixel offset - resulting in a cleaner preview (should no longer result in showing parts of a subsequent/previous frame) 🎉

main-support.ts
```diff
@@ -590,17 +578,28 @@ function extractMetadataForThisONEFile(
       const origWidth = metadata.streams[0].width;
       const origHeight = metadata.streams[0].height;
       const sizeLabel = labelVideo(origWidth, origHeight);
-      const fileSize = metadata.format.size;
+      const fileSize: string = metadata.format.size; // looks like a number, but actually a string!
```
cal2195: Is this intentional? ie. is there a reason we have it as a string?

main-support.ts
```diff
@@ -584,17 +580,28 @@ function extractMetadataForThisONEFile(
...
   });
 }
 
+/**
+ * Hash a given file using it's file name and size
+ * md5(file.name + file.size)
+ * @param fileName  -- file name to hash
+ * @param fileSize  -- file size to hash
+ */
+export function hashFile(fileName: string, fileSize: string): string {
```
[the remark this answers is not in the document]
cal2195: Just FYI - I'm thinking about slightly tweaking the hash function! 👍

whyboris:
Sounds good -- feel free to edit the hash as you see fit - no worries about breaking changes as we work before v2.0.0 is released 👍

main-support.ts
whyboris: It turns out that `ffprobe` outputs `format.size` as a string (I checked with `console.log(typeOf(metadata.format.size))` -- so I added the type definition to track what happens to it 👍

cal2195: Ha! - Should we parse the integer here?

whyboris: Makes sense -- will do 👍

PR #49, opened 2018-12-28
Convert arrays to objects!

cal2195:
This converts arrays in the vha file to objects! Makes the code a lot easier to read! 👍 

Hopefully I haven't missed anything! 💯


cal2195:
I've also commented out the demo content as it was not compiling anymore! 👍

whyboris:
I love it! Works beautifully!

Needless to say old files don't open, but we're breaking backwards compatibility ✅ so no worries!

PR #52, opened 2018-12-29
Statistics menu

whyboris:
Show basic statistics. In the future we'll have some fancier stuff there.

The _size_ will show up correctly once #48 lands (it corrects the _fileSize_ field in the `vha` file from _string_ to _number_)


whyboris:
Confirming that file size _total_ and _average_ will work once #48 lands 👍

PR #54, opened 2018-12-30
4K filter now works

whyboris:
4K filter didn't work because it was looking for `4K` not `4k` (capitalization matters).

Future `.vha` files will have correct string when extracting.

Fix includes strongly typing `resolutiongString` everywhere in the app 👍 

bonus: resolution in _file-view_ now aligned right (prettier)


whyboris:
@cal2195 -- feel free to skip reviewing this -- I'm just defaulting to showing you every PR I make in case you're curious. 

This PR is an example of using _TypeScript_ to avoid the kind of bug I created when I wasn't careful before 😅

Issue #50, opened 2018-12-28
Horizontal filmstrip bug tracking


whyboris:
The last PR #44 resulted in minor bugs when extracting videos, this PR does not fix them.

*Thumbnails **don't** show up* if:
- [x] the `vha` file is in the same directory as source files
- [x] the `vha` file has a space in its name
- [x] the screenshot size is 50px (and likely other sizes) - due to rounding error.
- [ ] Sometimes a slice of the next frame still appears in the thumbnail view
<img width="303" alt="screenshot 2018-12-29 at 15 40 32" src="https://user-images.githubusercontent.com/9119112/50539909-2c644900-0b80-11e9-938e-63b12560c46e.png">

- [x] The last frame is still often black in the film strip.
- [ ] Zooming out after scrolling to the end of the filmstrip prevents you from scrolling back when zoomed out!

This issue is meant to track various problems of this sort.

cal2195:
Some other little issues:

1) Sometimes a slice of the next frame still appears in the thumbnail view
<img width="303" alt="screenshot 2018-12-29 at 15 40 32" src="https://user-images.githubusercontent.com/9119112/50539909-2c644900-0b80-11e9-938e-63b12560c46e.png">

2) The last frame is still often black in the film strip.

3) Zooming out after scrolling to the end of the filmstrip prevents you from scrolling back when zoomed out!

cal2195:
Thumbnail preview and filmstrips aren't displaying for me on this MacBook - the video previews are though.

They're being generated, but the background-image: url isn't appearing in the final HTML. It worked last night on another system - could be some caching issue.

Still investigating!

cal2195:
Ha, nevermind! Realised the issue is one you've already reported, although slightly expanded:

_Thumbnails **also don't** show up if:_
- The path to the `vha` file has a space in it!

I think you're first point is incorrect though - I have a vha in the same location as the source files and it works fine?

👍

whyboris:
I'll look into this, probably close this issue and reopen a clean one with just the bugs that still persist 👍

whyboris:
#131 <-- tracking for this particular issue 👌

whyboris:
Closing -- we should open a new ticket in case something (re-)appears 😅

PR #58, opened 2019-01-02
Implement scanning for file list only once!

### LINKED ISSUE #21: App needlessly scans for initial video list twice
When you first select a video directory, the app scans for all the videos, but just counts them and throws the list away.

Then when you click import, it scans them all again.

This could be improved by only scanning once and saving the result for the import!

### LINKED ISSUE #57: Initial file count should also ignore vha thumbnail folders!
Similar to #3 and #24, the initial file count should ignore `vha-*` folders!

cal2195:
This PR fixes #57 and also fixes #21! 🎉 

On a side note, thumbnail preview and filmstrips aren't displaying for me on this MacBook - the video previews are though. 

They're being generated, but the `background-image: url` isn't appearing in the final HTML. It worked last night on another system - could be some caching issue, could you try a fresh build on your end? 

I don't think it's related to these changes. I'll investigate some more!

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -387,7 +387,7 @@ ipc.on('choose-input', function (event, someMessage) {
   }, function (files) {
     if (files) {
       const inputDirPath: string = files[0];
-      event.sender.send('inputFolderChosen', inputDirPath, numberOfVidsIn(inputDirPath));
+      event.sender.send('inputFolderChosen', inputDirPath, getVideoPathsAndNames(inputDirPath));
     }
   });
 });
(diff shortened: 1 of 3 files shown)


cal2195:
I discovered the issue - see #50! 👍

whyboris:
Thank you for the PR -- I didn't get a chance to review today -- will do tomorrow after work ✅

whyboris:
We can now safely delete the whole method `numberOfVidsIn` 🔪 

Great improvement 👍

Issue #13, opened 2018-12-12
Optionally include folder names in file tagging algorithm


cal2195:
This can be useful if a user has already organised videos by folders, eg. movie genre.

Adding the file path (minus any path shared by all videos) into the tag generation would improve tagging accuracy.

cal2195:
I've tried this out and it's very easy to implement without many changes! 👍  It also adds a lot more context to videos that are sorted well by folders.

One issue though - it basically turns the `file filter` into the `magic filter` as you're now filtering by both files and folders.

So, do we:

1. Only show a generic filter (files and folders) when this feature is enabled.
2. Track what tags are from folders, and add them to the folder filter instead. (How do you deal with tags that are present in both? Add them to both filters?)
3. Something else I haven't thought of.

😕 

PS: It might also be nice to display folder tags separately in the full tag modal, so you can easily remove tags that mark too many files (eg. a root folder than contains a lot of files, but doesn't add much value).

whyboris:
Sorry I hadn't joined in on this deliberation before. Thank you for the PR #64 that improves part of this.

I don't have strong opinions here. I'm tempted to keep a "file only" search for only files. I'm unsure how others use the app, but I meant the "magic" filter to be the catch-all that might be enough for most people.

I also wonder if once we implement manual tagging some of these issues will go away?

cal2195:
Thanks for some inspiration! 😄 

After some careful thought, I propose the following new search UI:

**_Sidebar_**
folder union - `prefix folder path union`
folder intersection - `prefix folder path intersection`
**folder exclusion** -`prefix folder path exclusion (missing currently)`
filename union - `full file name union (including extension & special chars)`
filename intersection - `full file name intersection (including extension & special chars)`
filename exclusion - `full file name exclusion (including extension & special chars)`
**tags union** - `tags only union - see below`
**tags intersection** - `tags only intersection - see below`
**tags exclusion** - `tags only exclusion - see below`
magic search - `as-you-type file & folder search (maybe tags too?)`

**_Tags_**
Tags can be optional composed of any combination of the following:
1) Auto generated filename tags - clean file names
2) Auto generated folder tags - clean folder names
3) Manually added tags - user added tags
4) File extensions - lowercase file extensions (would be useful for #60)

Tags which are enabled will also affect what shows up in the word cloud. Clicking on a tag will add it to the **tag** fields instead of the file fields.

[...]

PR #62, opened 2019-01-02
Fix to only show meaningful tags!

### LINKED ISSUE #14: Remove tags already selected from tag cloud
It's pointless to display tags already selected in the tag cloud, as clicking on these do nothing.

It also pollutes the tag clouds, as selected tags always appear first as they are present in all videos filtered by those tags.

cal2195:
This PR doesn't show any tags that are present in all displayed results!

This fixes #14, and also means helps remove redundant parent folders when #13 is implemented! 👍

### DIFF
--- a/src/app/components/pipes/word-frequency.service.ts
+++ b/src/app/components/pipes/word-frequency.service.ts
@@ -69,12 +69,12 @@ export class WordFrequencyService {
    * remove it from the map,
    * and return it with its frequency as an object
    */
-  private getMostFrequent() {
+  private getMostFrequent(total: number) {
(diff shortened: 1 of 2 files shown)


src/app/components/pipes/word-frequency.service.ts
```diff
@@ -69,12 +69,12 @@ export class WordFrequencyService {
    * remove it from the map,
    * and return it with its frequency as an object
    */
-  private getMostFrequent() {
+  private getMostFrequent(total: number) {
```
whyboris: We'll add the parameter to the _JSdoc_ later 👌

PR #64, opened 2019-01-02
Add support for importing new files, and verifying thumbnails!

cal2195:
This PR is the start of #4 - it can be merged but I'm planning on refining it more on the UI side.

It adds support for finding and importing new files _only_, and also for verifying all files have thumbnails (resumable thumbnail importing).

I'm not happy with:
- The button labels: I'd prefer to have an action, and a description underneath. For example:
```
[] **Import New Files**
This will scan for new files only and import them, leaving current files alone.
```
- There's still no _full_ rescan option - I need to think if it's needed, or if the optimised rescan is enough!

- I'd like these function to be pinable to the button bar! (needs new icons)

- The strings need to be translated! 👍

### DIFF
--- a/src/app/i18n/ru.ts
+++ b/src/app/i18n/ru.ts
@@ -35,9 +35,14 @@ export const Russian = {
     randomImageHint: 'Показать случайный скриншот',
     randomizeGalleryDescription: 'Рандомизирует порядок видеофайлов после каждого поиска',
     randomizeGalleryHint: 'Рандомизировать порядок галереи',
+    // TODO: Translate these
+    importNewFilesDescription: 'Quickly scan for new files only and import them',
+    importNewFilesHint: 'Import new files',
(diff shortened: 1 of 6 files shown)


src/app/i18n/ru.ts
```diff
@@ -35,9 +35,14 @@ export const Russian = {
     randomImageHint: 'Показать случайный скриншот',
     randomizeGalleryDescription: 'Рандомизирует порядок видеофайлов после каждого поиска',
     randomizeGalleryHint: 'Рандомизировать порядок галереи',
+    // TODO: Translate these
```
whyboris: Thank you for the `TODO` -- I'll get these translated.
My dirty trick is that I simply used Google Translate :trollface: 
Though I am fluent in Russian so it's easy to confirm the translations aren't bad 👍

whyboris:
Sorry didn't have time to review today -- will get it done tomorrow ✅

PR #70, opened 2019-01-04
Optimise output resolutions for scaling!

### LINKED ISSUE #67: Performance Issue: Output Resolutions not multiples of 16
As seen [here](https://ffmpeg.org/pipermail/ffmpeg-user/2017-October/037501.html), for optimal scaler performance, the output resolutions should be multiples of 16.

We should really stick to 16:9 ratios like these:
240p, 360p, 480p.

This will also help with the encoding process! I've noticed a large slowdown since scaling was introduced!

cal2195:
This PR fixes #67, switching to true 16:9 resolutions that are divisible by both 8 and 16, improving performance, fixing rounding issues/off by one pixel errors (as the width and heights are always whole numbers), and supporting more codecs.

I got the list of resolutions from here: https://pacoup.com/2011/06/12/list-of-true-169-resolutions/

Feel free to use different ones if you'd prefer, any of the divisible by 8 & 16 ones are good! 👍

We could also change the labels displayed to the user if they look too confusing, but keep the actual values!

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -221,8 +221,8 @@ export function generateScreenshotStrip(
   let allFramesFiltered = '';
   let outputFrames = '';
 
-  // Hardcode a specific ~16:9 ratio
-  const ssWidth: number = Math.ceil(screenshotHeight * 1.78);
+  // Hardcode a specific 16:9 ratio
+  const ssWidth: number = screenshotHeight * (16 / 9);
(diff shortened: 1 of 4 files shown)


whyboris:
Thank you! I was trying out previously with variations but I should have done my math better - my attempts included a height of 99px (since 99 / 9 works) - I also tried things that were divisible by 2 as well, but I see how divisible by 8 makes things more robust 👍

Issue #53, opened 2018-12-29
Manual Tagging!


cal2195:
This issue is to discuss and implement the much requested feature - manual tags! 💯 

Some implementation ideas I've had:

1) We can add a `tags:` attribute to the `ImageElement` object to save the tags with each file.

2) Special tags can be used for "Starred" videos etc. They can piggy back off this with a tag of `starred`.

Ideally it should be as easy as possible to add many tags, to many videos!

Feel free to add additional ideas/info below! 👍

whyboris:
Great to have a public place to share thoughts on this large feature.

The biggest hurdle for me with tags has always been the _UI_ (user interface) -- adding tags to an _ImageElement_ in code is super simple; so is searching through tags. But getting a clean interface / process for adding tags is not obvious.

Some features that would be nice to have:
- auto-complete (in case a tag already exists, suggest it)
- drag-drop (move tag onto a video would add the tag

UI questions:
- removing tags (when / where / can you remove a tag from all videos or just one at a time?)
- viewing list of all available tags (maybe a horizontal tab at the bottom of the gallery?)
- adding tags (dragging? typing? can you apply a tag to all results shown on screen?)

Possible solutions:
- a special "edit tag" view (in addition to _filmstrip_ and _thumbnail_ views) - it would show all videos as the _thumbnail_ view does, but only show one video per row. The folder/title of the video would be on the right side of the thumbnail, as would be all the tags and other metadata (possibly "number of times viewed", "star rating", etc).

I'm happy to implement the "edit tag" view first as a place where all the tags can go (along with a rudimentary "add tag" field). Adding a filter (search) on the left bar that searches through tags is super simple (can be first done without auto-complete). All other pieces (and solutions to problems) can be added later.

Any thoughts on this rough outline of how to get started?

cal2195:
That sounds like a very good initial implementation! 👍 We can improve things later on anyway!

One feature I'd really like is the "tagging all results on screen" - I'm already planning on adding features to move/delete/mark all files on screen, so being able to tag them too would be very nice!

One use case for this would be to search for all the files you want to move/delete using the excellent filtering system, tag them as "to delete", do it a few more times! Then show all "to delete" tagged files, and delete them!

It would also allow for tagging many files, that may be under a different name you want to change!

cal2195:
Not sure how you're planning on implementing the tags, but this seems very flexible and has autocomplete/drag and drop features! 💡 
[...]

PR #73, opened 2019-01-04
Implement Tag Filters!

cal2195:
In preparation for #53 and #13, this PR adds 3 new filters for tags! 🎉 

Clicking on a tag now adds it to the tag filter, rather than the file filter! 👍 

To see one of the differences between the filters, try search for a file extension in both the file and tag filters! 💯 

This will become a lot more useful once manual tags, folder tags and extension tags are added! 👍


src/app/components/home/home.component.ts
```diff
@@ -650,15 +650,15 @@ export class HomeComponent implements OnInit, AfterViewInit {
...
-        this.settingsButtons['exclude'].toggled = true;
+      if (!this.settingsButtons['tagExclusion'].toggled) {
+        this.settingsButtons['tagExclusion'].toggled = true;
       }
-      this.onEnterKey(filter, 4); // 4th item is the `file` exlcude filter
+      this.onEnterKey(filter, 7); // 7th item is the `tag` exlcude filter
     } else {
-      if (!this.settingsButtons['file'].toggled) {
-        this.settingsButtons['file'].toggled = true;
+      if (!this.settingsButtons['tagIntersection'].toggled) {
```
whyboris: The `handleFileWordClicked` method gets called when a word in the file name in the top bar is clicked - so the old behavior (opening `file` search) should be maintained in that scenario: https://github.com/whyboris/Video-Hub-App/blob/master/src/app/components/home/home.component.html#L402

This is because I intended for users (and suspect the current ways users use the feature is) to click on a word in the file name and see other files that have that word.

Once _manual tags_ are added, and the tag search (_tagIntersection_) will be searching manual tags, the intended functionality will disappear.

Similarly the 'word cloud' currently is correlated with the file name not with the tags.

Perhaps we just make `handleFileWordClicked` include another parameter (say 'tags: bool' -- and depending on whether the parameter is true or false, it will search either file names or tags?

Also unlike before this change, the 'shift' key doesn't seem to affect when clicking on a word in the file name.

whyboris:
It's looking great. We'll definitely merge this in within the week. I just have the small concern over `handleFileWordClicked` (see comment) - but that could be fixed rather easily.

I'll try to get a very basic version of tags as a PR and after your approval add into the _master_ (hopefully I can finish this weekend). And then this set of filters will hook into the field it's intended to search 🎉

src/app/components/home/home.component.ts
cal2195: Haha - I hadn't even realised that this was functionality! 😂 Are you sure 'shift' click worked before hand? I don't think the event gets passed through!

And I think we're slightly on different pages here - the idea of the `tag` filters are that they are customisable. As I discussed in #13, my idea is that you can have any combination of the following included in tags:
- clean file names
- clean folder names
- manual tags
- other file metadata (extensions etc.)

So since the top file name words are already cleaned, it makes sense to pop them in the tag filter if you have `clean file names` for tags enabled! Since they are the same, the functionality won't change.

Say you just enable manual tags for tagging, then we can have it pop top bar words into the file search instead - but I would prefer to do this once tagging is initially implemented! 👍 

Let me know if you don't get what I mean - want to be on the same page going forward! 😄

cal2195:
Looking forward to the basic version of tagging! 👍 👍

src/app/components/home/home.component.ts
cal2195: Just in case you've missed it - what do you think of my thinking above? 😄

whyboris: Sorry it took so long to get back to your question / idea / proposal -- busy week.

I feel like I might be misunderstanding you. I'll try to summarize what I think you're recommending and you can correct whatever misunderstanding I have. 

Part of the communication problem may be that I hadn't coherently labeled what things are called. So in the text below I'll try to be more clear.

- *auto-tags* - the tags that are automatically extracted from the file names (auto-capturing two-word combinations like "Batman Returns" if you have many files that include that word combination).
- *manual-tags* - the feature not yet in _master_ but that I'm working on at the moment - allowing users to add additional string descriptions to any file.
- *filter* - a user-editable field in the sidebar that removes from view from the full list of videos items that don't fit some rule (e.g. don't contain the word "batman").

In #13 I think you're proposing that when the app generates the *auto-tags* list it _also_ include folder names on the auto-generated list (possibly colored differently; perhaps if the folder contains more than 5 videos inside it). If that's done, then clicking on any one of the entities should pop the folder name into a *filter* on the sidebar; but at the moment there isn't a place which would accept both a folder name and a file name - hence the need for some hybrid filter (other than the "magic" filter we already have).

Is this roughly correct?

cal2195:
@whyboris If you're happy with my idea for now, this can be merged! 👍 

Next steps are:
1) Make it when you click on a manual tag, it get added to the new tag filter.
2) Optionally show auto tags in the detail view.

whyboris:
Will try to find time to review today when I'm home from work 👌

PR #74, opened 2019-01-04
Add a script to jump start new translations!

cal2195:
✨ _A wild script appears!_ ✨ 

A little script that jumpstarts new translation files! 🎉 

To use it, call it like this:
```
./translate.sh <const name> <file name>
```
For example:
```
./translate.sh French fr
```

This will create `fr.ts`, strip out all the English strings into a file called `to_translate.txt`.
You then copy the entire file contents into Google Translate, and paste the results into a file called `translated.txt`.

Hit enter, and the script will take your translations and replace the English strings in the new file with them! 🎉 

It's not perfect - Google gives back slightly different results than if you did it line by line - but it serves as a quick way to jump start many languages! 👍


whyboris:
That's fantastic! 

There's also https://translate.google.com/toolkit/docupload which lets you upload a document rather than copy-paste its contents 👍 

The biggest hurdle for language translations is bringing the translations back into the original `.ts` file, but you solved it 🎉 🎉 🎉 

It works almost perfectly -- just seems to have trouble with the `&` character:
`galleryAndView: 'Gallery & view settings',` becomes
`galleryAndView: 'Gallery 'Gallery & view settings' view settings',`

cal2195:
Thanks Boris for catching that! 👍 

I've fixed it by using node now instead for literal string replacement that works a lot better! 🎉 

That docupload site is awesome! I've added in French with this as an example too! 👍

whyboris:
Almost perfect!

Now seems to add a new line character before the `',`
```
export const English = {
  BUTTONS: {
    clearHistoryDescription: 'Clears the atoeuhasonetuhasentuhad history
',
    clearHistoryHint: 'Clear anoethuasnoethuasoth history
',
```

whyboris:
The Windows problem I had was an extra new line character getting added -- I suspect it has something to do with how Windows and Mac use different `CRLF` vs `LF` line ending. Merging it in 👍

Issue #31, opened 2018-12-19
Cmd+Q on Mac doesn't trigger vha file resave


cal2195:
If you use `cmd+q` on Mac to quit the app, the vha file doesn't get resaved if required. (Eg. a file renamed) 😢 

The correct behaviour occurs when you quit using the `x` in the window title bar. ❌

cal2195:
Looking a bit more into this, it looks like we need to implement:
```
app.on('before-quit', (event) => {
});
```
And use it to call `initiateClose()` in `home.component.ts`!

Seems like you can't catch `cmd+q` with javascript anymore - also that key combo is there twice for two different functions:
```
// Listen for key presses
  @HostListener('document:keydown', ['$event'])
  handleKeyboardEvent(event: KeyboardEvent) {
    // .metaKey is for mac `command` button
    if (event.ctrlKey === true || event.metaKey) {
...
      } else if (event.key === 'q') {
        event.preventDefault();
        event.stopPropagation();
        this.initiateClose();
      }
...

      } else if (event.key === 'q') {
        if (!this.settingsButtons['magic'].toggled) {
          this.settingsButtons['magic'].toggled = true;
        }
        this.showSidebar();
        setTimeout(() => {
          this.magicSearch.nativeElement.focus();
        }, 1);
      }
...
```

whyboris:
Thank you for finding the reference!

https://github.com/electron/electron/blob/master/docs/api/app.md#event-before-quit

Looks like we can just `preventDefault()` in `main.ts` and let the electron front-end handle it 👌

whyboris:
Should be fixed here: https://github.com/whyboris/Video-Hub-App/pull/79

PR #79, opened 2019-01-05
handle Mac cmd + Q not saving setting bug

whyboris:
🐎 fix that bug!

Thank you @cal2195 for catching it and for suggesting the fix in the issue.

This PR will close https://github.com/whyboris/Video-Hub-App/issues/31

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -145,6 +145,10 @@ try {
     // }
   });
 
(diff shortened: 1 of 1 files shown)


cal2195:
Nice!! 🎉 I tried to fix this myself, but I couldn't get it working!
Thanks for that! 👍

PR #82, opened 2019-01-05
Instructions for adding a language

whyboris:
And integrating French into the app 👍  Thanks @cal2195 for the script and for the French translation!

### DIFF
--- a/src/app/i18n/README.md
+++ b/src/app/i18n/README.md
@@ -1,0 +1,47 @@
+# Adding a language
+
+Thank you for considering adding an additional translation to Video Hub App!
+
+It's a 2 minute process when using the script (created by @cal2195) and [Google Doc Translate](https://translate.google.com/toolkit/docupload). Or if you are translating individual strings you can use [Google Translate](https://translate.google.com/).
+
+## Steps
+
+1. In your terminal enter this directory: `cd src/app/i18n`
+2. Run the translate command: `./translate.sh German ge` substituting the language and the two-letter abbreviation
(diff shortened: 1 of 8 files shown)


src/app/i18n/README.md
```diff
@@ -0,0 +1,47 @@
...
+3. The instructions in the script will inform you:
+> copy to_translate.txt contents into Google Translate, and paste the translations into a file called translated.txt
+4. Once done, press the `Enter` key in your terminal and you're done!
+5. Open a PR (Pull Request) with the changes.
+
+## Integrating the language into the app
+
+You're welcome to open a Pull Request as is and I'll implement the language into the app. Alternatively, you can edit just a few more files and you'll be ready! It's rather eays -- just follow the patterns!
+
+1. The newly generated file will need a change: `export const English = {` -> to the appropriate language name
```
cal2195: This is supposed to happen with the script - does it not?

whyboris: It _should_ happen in the script 😅 -- I think it doesn't -- checking again now

whyboris: ps - I also removed all the `tslint` rules from within the file -- and I'm keeping only one rule at the top of the file. Might make the script simpler (if you're editing it).

cal2195: It's the scripts fault! I'll fix! 👍

whyboris: Seems to not put the language name into the `ts` file still - but it's likely not worth the effort to fix 👌

PR #83, opened 2019-01-05
Fix for corrupt metadata in files!

cal2195:
It appears in some files, the metadata for duration is longer than the actual streams themselves! 👎 This causes the thumbnail generation to fail as it tries to seek to timestamps that have no video!

This fix looks at both the metadata, and the stream duration, and chooses the shorter of the two! 👍


cal2195:
I've changed it to prefer the stream duration, as that should be always correct - unless it's 0, in which case it will pick the format duration (damn mkv files).

PR #87, opened 2019-01-05
fullView - view every screenshot of every video at once in one view!

whyboris:
View the entire filmstrip at once: vertically stacks enough filmstrip bars to show the full video preview.

Please note any bugs you notice. Should work just fine 😅 

Cheers 🎉


whyboris:
@cal2195 - if you merge this PR, please `Squash and merge` so that it goes into the history as a single commit 😅

cal2195:
Getting consistent crashes will this, stack traces and all! 😕 I'll see if I can figure out why...

cal2195:
Zoomed far out, it works, but only uses a fraction of the width. Zoomed in it crashes before displaying anything! Might be some kind of divide by 0 maybe?

I'll investigate more soon!

whyboris:
Thank you for discovering problems with my implementation. I'll spend some time today debugging (was up late in the evening whet I finished, should have tested more edge cases) -- I basically never use the zoom feature 😅

cal2195:
Found it! You forgot to update the gallery width on zoom!
```
    } else if (uniqueKey === 'makeSmaller') {
      this.decreaseSize();
      this.updateGalleryWidthMeasurement();
    } else if (uniqueKey === 'makeLarger') {
      this.increaseSize();
      this.updateGalleryWidthMeasurement();
    } 
```
Is there a way we could hook into the size instead of remembering to add this everywhere?
It was a divide by 0 that was causing the crash! I added a check to be safe before continuing if it's not 0! 👍 

Two more bugs:
1) You have to zoom before the full display correctly computes the size (maybe there's a better way to do this?)
2) I'm noticing a lot of jumpy scrolling - like things are popping in and out when I'm scrolling in this view - maybe the scroll view is computing the incorrect size of the items?

Looking good overall though! 💯

whyboris:
Thank you for describing the two bugs. I'm aware of both. Both just require a bit of my time (I think I finally have some today).

_ps_ When I was making the app I often took shortcuts - writing code fast with little regard for future improvements. Hence there's lots of room to improve and streamline. This PR is another example - I just peppered `updateGalleryWidthMeasurement()` everywhere and called it a day 😓

whyboris:
Thank you again for catching the _divide by 0_ bug. 
I know this isn't 100% polished, but I think it's good enough at the moment and I'm really eager to get the _tags_ into the app so as to no longer block your other PRs.
If you think it's a decent first draft - please merge it in 🙇 
I'll work on creating the _tags_ PR and later on once we figure out the better zoom system (one you recommended) I could work on this feature - to make not jump, etc 👍

PR #88, opened 2019-01-07
Default screenshot size bug fix!

cal2195:
Found a bug, where if you didn't change the default screenshot size, it would import at 200px - worst case failing to import certain video resolutions! 👎


whyboris:
Great catch 🎉

Issue #11, opened 2018-12-12
Filter by video length


cal2195:
Just as you can filter videos by video resolution, it would be nice to be able to filter by video length as well.

Both an upper and lower bound would be good! 👍

whyboris:
I'll work on getting this in 👍

cal2195:
Have you started on this? 😄

If not - I wouldn't mind looking at this! 👍

whyboris:
Haven't started yet -- feel free to give it a go.

I may be a bit finicky about the UI - how were you thinking of letting users choose the cutoffs?

ps - there seem to be two features, both we'd like to have eventually:
- filter by length using minimum and maximum time cutoffs
- _sort_ by length - ascending / descending (separate issue from this) 👍

cal2195:
There's lots of sorting that would be nice! 👍 We should make the issues for them! (Maybe a parent issue, and one for each type)

And I was planning on copying the resolution filter, with 0->max_video_length, and some number of divisions in between... Haven't thought too much about it - have you any ideas? 💡

whyboris:
We can easily detect the shortest and longest video within the hub - set those as cutoffs, and then have the two sliders as I have with the resolution filter.

What made me pause is that I custom coded the resolution filter sliders but not in a flexible-enough way that I could reuse the component. I've not looked into it in more than half a year -- perhaps it would be easy-enough to modify. Do take a look if it can be!

For the UI, I suspect a nice way to do it would be something like this:

```
above:          15min           45 min
bar:         ======0####0=============
below:     3min                                    1hr 24min
```

So that the min and max are always shown (below) and the cutoff is auto-updated as you slide the slider just above the div that you drag (minimum aligned left, maximum aligned right - so they never overlap even if close together)

Comments? 😃

cal2195:
Added in #112!

PR #90, opened 2019-01-07
Converted ResFilter into a generic SliderFilter!

cal2195:
In preparation for #11, I've converted the ResFilter into a generic, reusable slider filter! 🎉 

I had to move about the resolution frequency graph display a little to match the slider mark points - feel free to make it prettier! 👍


whyboris:
Looking forward to trying it out! Looks good from the code perspective 💯

PR #93, opened 2019-01-08
Added version to vha file! Bumped to v2.0.0a!

### LINKED ISSUE #9: Add versioning to the vha file
Without a version field in the vha file, it's hard to work out what type of vha you are opening, and how to convert it to a new version.

We should add a version field, so that we can add backward compatibility properly. (We can assume no version means <=v1.3.0).

cal2195:
This PR closes #9!

It adds version info to the vha file, and also bumps the version of the app to v2.0.0a! 👍

### DIFF
--- a/main-globals.ts
+++ b/main-globals.ts
@@ -1,4 +1,6 @@
 export const globals: Globals = {
+  version: '2.0.0a',
+  vhaFileVersion: 2,
   angularApp: null,               // reference used to send messages back to Angular App
   cancelCurrentImport: false,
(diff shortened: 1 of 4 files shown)


whyboris:
Looks good - we'll just reset the counter to version `2.0.0` when releasing to the public.

Just confirming that you'd like this behavior: every subsequent release of the app will bump the version of the `.vha` file even if there are no interface changes?

cal2195:
Hmmm, you make an interesting point - maybe fixed "file format" versions would be a more sane choice... 🤔 💭

cal2195:
Implemented your idea! Fixed vhaFileVersions! 👍

Issue #71, opened 2019-01-04
Snap zoom levels to use full space available


cal2195:
@whyboris 
Wanted to get your opinion on this! 👍 

I don't really see a point in zooming arbitrary levels (+-25px I think atm), since most of those won't use the full space of the viewport.

I would like the zoom to change the number of files displayed in one row, filling the entire row - what do you think of this?

cal2195:
You can see how this would work in #72! 👍

cal2195:
I have this feature under way! 😃 Will also bundle in some overall zooming/view improvements! 👍

whyboris:
Thank you Cal for #170 -- it implements the solution to this issue beautifully 🙇

PR #95, opened 2019-01-08
Fix to remove black at end of filmstrip preview!

cal2195:
This PR fixes one of the filmstrip bugs in #50 - it removes the black at the end of a filmstrip! 🎉


whyboris:
I'll try it out later this week 👍

cal2195:
An alternative is, along with #71, we could have the filmstrip only show an integer number of screenshots with no cut off - what do you think of this? 💡

cal2195:
I've implemented my last comment in #170! 👍

cal2195:
This has been fixed in #170 - Closing! 🎉

PR #105, opened 2019-01-14
Star rating implemented

whyboris:
First draft but it works.

The stars will later be actual gold ⭐️ ⭐️ ⭐️ rather than the current _VHA_ icon.

I'm open to suggestions -- 3 star max or 5 star max? Should be very easy to edit.

I'm not proud about how I mostly copy/pasted the _resolution filter_ setup, rather than keeping my code _DRY_. But I'd rather not optimize ahead of time -- we'll likely be adding more filters, and it will be easier to see later how to consolidate them, rather than spend time now over-optimizing with _n=2_.

Cheers!

@cal2195 - please merge unless you see bugs or recommend I change something 🙇

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -155,6 +155,7 @@ export function getVideoPathsAndNames(sourceFolderPath: string): ImageElement[] 
                 numOfScreenshots: 10, // hardcoded default
                 partialPath: partialPath,
                 resolution: '',
+                stars: 0
               };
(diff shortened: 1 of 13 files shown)


cal2195:
You can either fix the bug first, or merge and fix later! 👍

whyboris:
Thank you for catching that bug. I'll do cleanup in a later PR 👍

PR #106, opened 2019-01-16
Tag tray - first draft

whyboris:
Currently: 
- new button next to the `dark mode`
- clicking it opens the tag tray (bottom)
- it lists current available tags
- 🎉 the list updates when you add new tags in the details view 🎉 

Later:
- will be able to show frequency of tags
- will be sortable (alphabetical or by frequency)
- clicking on a tag will add it to the _tag_ search


whyboris:
So much to do, so little time! 😅 

@cal2195 - merge at will 👍

cal2195:
We can do it! 👍

PR #107, opened 2019-01-17
Tag tray updates

whyboris:
- can filter tags in tag tray via text input
- tags in tag tray sorted alphabetically or by frequency - based on user setting (new button)
- star rating now shows up correctly
- tag tray allows you to show or hide the frequency of each tag 🎉 
- 😁 

⚠️ the button to show/hide frequency is just an ugly placeholder - but please don't let that block merging this branch if everything else is ok 👌

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -46,6 +46,7 @@ import { FolderArrowsPipe } from './components/pipes/folder-arrows.pipe';
 import { FolderViewPipe } from './components/pipes/folder-view.pipe';
 import { LengthPipe } from './components/pipes/length.pipe';
 import { MagicSearchPipe } from './components/pipes/magic-search.pipe';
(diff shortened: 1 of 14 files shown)


whyboris:
Every PR I've been making is rough -- I'll be doing a pass later to make the look feel polished. Before releasing 1.0.0 I spent so many days messing with things a pixel at a time sometimes 😅 

For now the input to filter tags in the tab tray is just shoved to the side. Do you have thoughts whether to have it on the far right (as is), in the middle (as in the auto-tags menu) or on the left (where the title is - which then would have to be moved elsewhere).

This is a rather inconsequential decision, but happy to hear feedback if you have a preference.

whyboris:
ps - please `Squash and merge` whenever a PR has more than 1 commit 🙇  otherwise a regular `Create a merge commit` is fine 🙆‍♂️

PR #108, opened 2019-01-19
Sorting options - sort by size / duration / etc

whyboris:
Sort by:
- default
- size descending
- size ascending
- time descending
- time ascending
- random

⚠️ BREAKING CHANGES -- made some changes to the generated `vha` file.

TODO: In a future PR, simplify the silly `shuffleTheViewNow` hack and confirm that the old shuffle button and keyboard shortcuts work.


whyboris:
Please `Squash and merge` unless you have any reservations 👍 

I'm now opting to pre- (and re-) computing the resolution string (and resolution sorting bucket) every time we open a _vha_ file. This simplifies the resolution pipe somewhat, but also provides an important new feature: `index`. We no longer have to worry about using the _hash_ as the unique reference to a file (in case a user has a duplicate file with the same size and therefore hash).

In the future we could refactor the `index` to be the unique reference when using the `openVideo` method.

main-support.ts
```diff
@@ -556,15 +597,16 @@ function extractMetadataForThisONEFile(
       const fileDuration = metadata.streams[0].duration || metadata.format.duration;
 
       const duration = Math.round(fileDuration) || 0;
-      const origWidth = metadata.streams[0].width;
-      const origHeight = metadata.streams[0].height;
-      const sizeLabel = labelVideo(origWidth, origHeight);
+      const origWidth = metadata.streams[0].width || 0; // ffprobe does not detect it on some MKV streams
```
cal2195: I'll look into this! 👍

src/app/components/common/final-object.interface.ts
```diff
@@ -19,9 +19,14 @@ export interface ImageElement {
   fileName: string;              // for opening the file
   fileSize: number;              // file size in bytes
   hash: string;                  // used for detecting changed files and as a screenshot identifier
-  numOfScreenshots: number;      // number of screenshots for this file
+  height: number;                // height of the video
+  index: number;                 // for the `default` sort order
```
cal2195: Does this all get saved, and reserved into the vha file? 😕 

I'd prefer if we had a second object, or some other way where we don't save the things we can compute to the harddisk! What do you think? 💡

whyboris: After I submitted the PR and went to bed I realized the simplest thing to do is just have the `main.ts` (Node code) strip out all the fields we don't need to save 👍 

I'll make that change right now 🚀

PR #109, opened 2019-01-19
Corrupt video detection

cal2195:
This PR adds corrupt video detection, and fixes removing obsolete thumbnails! 👍 

This doesn't fix wmv files yet, but I will work on that! 💯

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -184,6 +184,63 @@ const spawn = require('child_process').spawn;
 // also spawns a shell (can pass a single cmd string)
 const exec = require('child_process').exec;
 
+export function checkForCorruptFile(pathToVideo: string,
+  fileHash: string,
+  duration: number,
+  screenshotHeight: number,
+  numberOfScreenshots: number,
+  saveLocation: string,
+  done: any) {
+    if (fs.existsSync(saveLocation + '/filmstrips/' + fileHash + '.jpg')) {
+      // console.log("thumbnails for " + fileHash + " already exist");
+      takeTenClips(pathToVideo,
+                   fileHash,
+                   duration,
+                   screenshotHeight,
+                   saveLocation,
+                   done);
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


whyboris:
Looks great. Seems to work.

One console error I saw repeated in _node_: 

```
[mov,mp4,m4a,3gp,3g2,mj2 @ 00000248aa0cd580] st: 0 edit list: 1 Missing key frame while searching for timestamp: 0
[mov,mp4,m4a,3gp,3g2,mj2 @ 00000248aa0cd580] st: 0 edit list 1 Cannot find an index entry before timestamp: 0.
```

Is this expected? It seems like it's looking for an index entry before _timestamp_ 0. Makes sense -- since before the first frame of the video is shown, the render process needs to have access to a key frame information. 

Just checking 👌

main-support.ts
```diff
@@ -184,6 +184,63 @@ const spawn = require('child_process').spawn;
...
+export function checkForCorruptFile(pathToVideo: string,
+  fileHash: string,
+  duration: number,
+  screenshotHeight: number,
+  numberOfScreenshots: number,
+  saveLocation: string,
+  done: any) {
+    if (fs.existsSync(saveLocation + '/filmstrips/' + fileHash + '.jpg')) {
+      // console.log("thumbnails for " + fileHash + " already exist");
+      takeTenClips(pathToVideo,
```
whyboris: Doesn't need to be done now, but what do you think about renaming `takeTenClips` to `extractVideoIfAbsent` (or something similar).

The method name makes it sound like it always generates a video with ten clips (rather than do it conditionally) 😅

whyboris:
@cal2195 -- you can _Squash and merge_ whenever ✅

main-support.ts
cal2195: Sounds good to me! 👍 I'll change it in the next one!

PR #111, opened 2019-01-19
Input & pipe cleanup

whyboris:
As more properties began to be added to the gallery elements, the _Inputs()_ list grew enormously.

This PR simplifies this by passing in the `video` (_ImageElement_) into each gallery display item, which in turns retrieves whichever property it wants (_resolution_, _duration_, etc).

Additionally, this PR removes the vast majority of the _timePipe_ and _fileSizePipe_ - to speed up the display when scrolling. The two display properties are now pre-computed every time a _vha_ file is opened (and removed whenever the _vha_ file is saved).

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -85,8 +85,10 @@ export function insertTemporaryFields(imagesArray: ImageElement[]): ImageElement
 
     // set resolution string & bucket
     const resolution: ResolutionMeta = labelVideo(element.width, element.height);
-    element.resolution = resolution.label;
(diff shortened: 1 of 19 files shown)


whyboris:
@cal2195 -- sorry I removed the `encodeURI` everywhere -- I'll put it back in. When I develop on _Window_ none of the images show up unless I remove this. I'll look sometime if it's easy enough to solve the problem once and for all - also allowing for spaces in _folder names_ 😅 

Ready for your review 👀

PR #112, opened 2019-01-19
Video length filter-ish

cal2195:
In this PR, I've added the beginnings of the video length filter! It filters correctly, but I'm having issues with the UI - I'm not as well versed in CSS! 😄 

@whyboris, Could you take a look at making the UI pretty?


cal2195:
I'll merge now and fix later! 👍

cal2195:
Easier to work off a common base! 😄

whyboris:
I'd like to move the time selection (visually and in processing order) before the _resolution filter_ - this would result in updates to the resolution histogram as we update the time selection.

I'll happily do it when I do clean-up. Or feel free to do it as you work, it's basically moving 1 line and 1 block of code up just before the above line and element 👍

PR #113, opened 2019-01-20
Icon overhaul

whyboris:
🎉 Massive success 🎉 

Except for the _folder-view.pipe_ which still uses the icon font, the rest of the buttons and icons all use the new `svg-definitions.component.html` file which simply has inline _SVG_ definitions!

As a bonus -- if we want to we can do multi-color SVGs (though will take a bit more code). By a happy accident the _tag_ icon has two colors 👍 

The _overwhelming_ reason behind this code change is to allow adding new icons just by defining a new svg. 

FYI - previously, all the icons had to be compiled with _FontForge_ (a bunch of correct dependencies had to be installed too) through a script, the resulting file and some code had to be manually copied. The script and dependencies only lived on my work mac which created added hassles. Good riddance 💥


whyboris:
I put a temporary placeholder for the _duration_ filter (which will be broken away from the _resolution_ filter later). I might not have time until Monday evening to finish that part -- so it's better to merge the PR if it doesn't have any major problems from your point of view 👌

PR #119, opened 2019-01-21
Index all the things two

### LINKED ISSUE #101: Files with identical hashes open wrong file
Minor issue with the new hash, should be easily fixed!

When you have duplicates of the same file, and they have the same hash, clicking on one to open it will opened the "last" version of the file! This is because `openVideo(hash)` takes a hash value, so it picks the last one with the matching hash, as they all share the same one!

To fix this, `openVideo()` should take the video path instead! 👍

cal2195:
This PR removes the fileMap, and references to imgId, and replaces them with index! 👍 

Fixes #101! 🎉

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -79,9 +79,7 @@ export function alphabetizeFinalArray(imagesArray: ImageElement[]): ImageElement
  */
 export function insertTemporaryFields(imagesArray: ImageElement[]): ImageElement[] {
 
-  let counter = 0;
-
-  imagesArray.forEach(element => {
+  imagesArray.forEach((element, index) => {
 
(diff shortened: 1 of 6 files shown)


main-support.ts
```diff
@@ -79,9 +79,7 @@ export function alphabetizeFinalArray(imagesArray: ImageElement[]): ImageElement
  */
 export function insertTemporaryFields(imagesArray: ImageElement[]): ImageElement[] {
 
-  let counter = 0;
-
-  imagesArray.forEach(element => {
+  imagesArray.forEach((element, index) => {
```
whyboris: 😅 Why didn't I do that 😓

PR #123, opened 2019-01-22
Show folder path in any gallery view + setting button to toggle it

### LINKED ISSUE #122: Display folder path in other views
If the user has the setting toggled, show the _folder path_ not just in the _file view_ but in all other views 😁 

Creating the issue for tracking - I have the feature first draft ready with _thumbnails_ view ✅

whyboris:
Show folder path in gallery (thumbnail, filmstrip, and other views, not just in _file_ view).

Ready for merging. For ease of review you can view only non-white-space-changes with `?w=1` flag, like this: https://github.com/whyboris/Video-Hub-App/pull/123/files?w=1

🐰 will close #122 

:trollface: the branch name was what I wanted to work on, but I ended up working on this instead 😅


cal2195:
This is so sweet! I love it! 😄 💯 

I'm seeing some issues in the filmstrip, detail and full view - unequal children sizes (folder vs item) is causing items to appear late/disappear early, and also jump the incorrect amount when scrolling! Any ideas on a fix?

whyboris:
Thanks for catching that -- I should have been validating my code against a hub that has more than 2 videos 😅 -- I'll see if I can fix it up this evening when I'm home 👍

cal2195:
Hahaha, sample size matters! 😉

whyboris:
Just tried out a hub with many folders in the directory and it's unacceptable 😭 

I'll see if I can find a solution 😅

whyboris:
Thank you for not merging! This was a mess 😢 

Turns out there's really no way around it: _virtual-scroller_ can't elegantly handle the complex gallery view we have when the heights of elements are different. So I opted for simply slapping the folder label above the first video in the folder.

Just needs some cleanup and it will be review ready 😅

cal2195:
Did you try out `[enableUnequalChildrenSizes]="true"`? If so, what was your experience?

From what I can see when I try it, it fixes every view, including your original implementation of this PR, and the full view, except the detail view! ❓ If that's the only view that doesn't work well, perhaps it's worth trying to fix that instead so we can make use of the much prettier design? 😄

cal2195:
I'm much more keen on your original implementation! 👍

cal2195:
I see - I wonder if there's any alternatives? I saw Angular 7 has it's own implementation, and they have an dynamic children height version in testing atm! Could be worth a look! 👍

whyboris:
I saw Angular 7 has its own. I'm worried it will take a lot of refactoring and debugging to get it to work. So far except for the _full view_ I don't expect any glitches from the setup we have.

But I'm open to it. Perhaps if I feel inspired I could give it a try -- might not be too hard to try out 😅

cal2195:
Trying to reproduce the issue you mentioned - when you use the unequal sizes, did you also remove the `bufferSize` and `childHeight` code?

All views are working for me without those getting involved?

whyboris:
Cool! I didn't remove the child heights 😓 I'll give it a try!

cal2195:
If you're happy with `enableUnequalChildrenSizes` without `bufferSize` and `childHeights`, feel free to revert to your first implementation and we can merge! 🎉

whyboris:
Thank you -- I'll try it out -- today I'm away to NYC all day. Will have time on Sunday 🗽

whyboris:
I got `Angular 7` Virtual Scroll to work (example: https://netbasal.com/a-taste-of-angular-material-virtual-scroll-f173c5c70a1)

It doesn't seem to handle more than one column of elements. I can make it display more items in a row with `display: inline-block` but then when elements are removed from the top (non-viewable area), they are removed one at a time - resulting in a horizontal shift of every displayed element by one. Unlike the current scroller which removes the whole row of items at a time, seamlessly transitioning as you scroll.

Perhaps there's a way to get it to work, but it's not instantly promising 😞

cal2195:
I've done some debugging and come to the same conclusion as you! I think you're theory is correct - I'm sure there's _some_ solution, but I'm not sure how easy/reliable it would be... 😕

cal2195:
#166 can probably fill in for this functionality!

whyboris:
Thank you Cal for #166 -- your implementation is better than I had intended 🙇

PR #125, opened 2019-01-22
Add details to sheet overview!

cal2195:
This PR adds star rating and tagging to the sheet modal! 🎉 

Future todo: reuse a component between the detail view and the sheet modal! 👍


whyboris:
I love that the _sheet modal_ is getting all the _details_! 🎉 

I bet you had this planned all along - yet it took me by surprise - an excellent addition 🗡

cal2195:
Haha, thanks! 👍 I was lazy so I didn't, but now I have! 😄 

I haven't refactored out the thumbnail component, but that would be a very good idea! 👍

cal2195:
Thanks for catching that! It broke in the refactor! 👍 Fixed!

PR #137, opened 2019-01-27
better menu labels and organization

### LINKED ISSUE #63: Support Multiple Headings in Settings Tabs
As far as I can see, right now you can only have a single heading in the settings tabs.

It would be nice, for example to have:

[Main Settings]
**Library Settings**
[] Quick Rescan
[] Full Rescan
[] Verify Thumbnails
**App Settings**
[] Start Wizard Again
[] Reset All Settings

Maybe this is possible already, but I couldn't see how to do it easily! 😄

whyboris:
Finally better labels for the menu.

Minor improvements and fixes.

⚠️ I removed _Angular_ animations from gallery items -- they slow down scrolling. They look pretty, but might not be worth the performance hit. Will re-evaluate before v2 release 👌 

Should close #63


whyboris:
You can _Hide whitespace changes_ for less clutter `&w=1` 😉 

I'm not too happy about including the menu section headings right in the buttons, but it makes things pretty easy to follow and move around.

I'm happy to hear proposals about clustering and organizing the menus differently 👍

cal2195:
I like having these headings! 👍  I don't mind that they're part of the button! 😄 

Three little things:
1. Do you think the buttons/text is too big now? I liked it smaller! (Albeit, I tend to use _a font for ants_ myself to get the most screen real estate! 😉 ) Maybe a little at least!

2. Instead of opening the 3rd tab for Main settings by default, I think we should move Main settings to the beginning! 👍 

3. Not necessarily in this PR, but what do you think of the idea of having drop down descriptions for some buttons?

whyboris:
I may have gone too big with the buttons -- I've been using a 4k 55 inch screen as my desktop (I sit further away) so I gravitated towards larger font. Part of my thinking was that having larger versions of the buttons might make them clearer for users (so it's easier to recognize them in the top ribbon).

The particular size I chose was a multiple of 2 for the actual icon resolution (as far as I can tell). 

I'm tempted to keep the UI minimal (which is why I went with a custom title bar: it's narrower than the standard one in Windows). But I wonder if the app settings should be as readable as possible. 

I'll try out on a 15 inch laptop to see how everything looks.

whyboris:
I wanted to put the _Main_ tab third so as to keep logical consistency: the buttons are all set in the same order - left-most tab correlates with the sidebar, etc.

But it does break up the convention a bit ... I don't have strong opinions on this 👌 I'll try it in the traditional order to see how it looks 👍

whyboris:
I'm happy to add helper text - explaining how a button works / etc.

To make the UI easier I might just render an `i` icon at the end of an explanation with text that pops up as soon as you hover over it. This will save users the need to click (and the temptation to click again to collapse it back).

cal2195:
Re font size: Could always make it change with the `larger font` option, so people can choose either! 👍 

Re tab order: It doesn't bother me too much - I hadn't noticed that the buttons matched how you add them to the bar, so it makes sense that way! 👍 

Re help text: What ever you think is easier, as long as it's easy to read a good bit of text if needed! 👍

whyboris:
I added an `i` icon next to _auto-tags_ to show what "more info" could look like. Feel free to add this information to any button that you think needs more elaboration than the standard description.

I tried to get a nice tooltip to show up when you hover over the `i` but I couldn't get the _z-index_ to make it go above the text of all the other descriptions. So I just made the text appear inline 🤷‍♂️

whyboris:
The font is still a bit larger than I suspect you'd like, but the buttons are back to their regular size. 

While the settings can be more compact, I'm trying to make sure that when the screen width is smaller, the text doesn't begin to overlap right away (currently it does - despite my attempts to do a `-webkit-line-clamp` hack. 

I'll revisit this topic again before v2.0.0 release - as part of a finishing cleanup 😅

whyboris:
I'm thinking I should move the `dark mode` button to the bottom of the menu (and the right in the toolbar) -- just before the `hide top bar` button 👍 

Will do in another PR 😸

PR #148, opened 2019-02-24
better default active buttons, reorder buttons

whyboris:
- reorder settings a bit
- reorder buttons a bit
- better defaults (most features hidden until toggled)


cal2195:
Hmmm, I like this, but I feel the current settings is a little hard to know what features do, and even how to find the ones you want! 😕 

Maybe we need to make it clearer, or implement modes to enable sets of features by default? (Basic, Advanced, Expert)

I personally turn everything on, but if I'm looking for a specific feature, I find it quite hard to actually find the toggle for it! 😆

whyboris:
I'm definitely going to be adding better icons, and I'm thinking about grouping the settings a bit better too. 👍 

all buttons

PR #152, opened 2019-02-25
Remember zoom level

### LINKED ISSUE #16: Remember Each View's Zoom Level
It would be a very nice addition to remember the last zoom level for each view, and they often look better with different zoom levels! 🔍

whyboris:
Remember each view's zoom label.

Closes #16 🎉 

⚠️  Almost ready -- but I'd like to clean things up a bit before merging. No review needed unless you'd like 👍

### DIFF
--- a/src/app/components/common/app-state.ts
+++ b/src/app/components/common/app-state.ts
@@ -2,14 +2,30 @@
 // https://github.com/electron/electron/blob/master/docs/api/locales.md
 export type SupportedLanguage = 'en' | 'ru' | 'fr';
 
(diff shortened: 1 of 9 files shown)


whyboris:
Might have minor bugs -- will fix those later 👌 

You might need to delete the `settings.json` where _Video Hub App_ stores its settings 🤷‍♂️

cal2195:
I love this! ❤️ 

I did need to delete my preferences though - which I think all users will need to do! ⚠️ 
(We should probably do it for them... and add versioning...)

cal2195:
There could also be a bug with this - every view except the first one (showThumbnails) works fine, but the first one doesn't seem to be restored on app restart! ❓

PR #160, opened 2019-04-04
Auto child sizes

cal2195:
This PR removes the fixed child size calculation, and uses auto child sizing instead, based on the first element of each virtual-scroller!

This also fixes the file element margins, so that their sizes are correctly calculated (a recent bug that had snuck in)! 👍


cal2195:
Closing in favour of #170!

PR #162, opened 2019-04-04
Nicer looking tooltips

cal2195:
Little change to the tool tips to allow for longer text, and make them look a little nicer! 👍 

Will follow up soon with more tooltips for the more complicated features!

### DIFF
--- a/src/app/components/home/settings.scss
+++ b/src/app/components/home/settings.scss
@@ -176,7 +176,6 @@ h3 {
   color: $black;
   font-size: 14px;
   left: 80px;
-  overflow: hidden;
   position: absolute;
   top: 2px;
 }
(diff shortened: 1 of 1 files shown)


whyboris:
It really is a matter of taste ...

Yours:
image

Mine:
image

Downside of mine is long text might begin to overflow (especially when the app width is small). Though I see it as a benefit of mine that the hint text is harmonious with the rest of the text (same style).

I would prefer to stay with my current implementation 🙇

whyboris:
I'm not really insisting on keeping the tooltip I currently have ... maybe if you share some of your thoughts on why a tooltip with a box may be better I might change my mind? :)

cal2195:
I agree that the style isn't very good - I think I was having some strange transparency/overlap issues - I actually wanted it to appear below, and even shift down the other buttons, but I couldn't get it to work! 😢 

Ideally, I just want to make it really easy to see what's help text - I think the current implementation is too hard to read when the text gets long, and the help text should be verbose to explain in detail each feature...

whyboris:
@cal2195 -- I'm aiming to close out tickets / PRs as we near the 2.0.0 release (few weeks from now earliest).

No rush on this -- just let me know sometime if I should reconsider the current tooltip -- perhaps to create a hybrid of the two.

I'm pretty good at CSS -- so you could sketch in Photoshop / some other editor what you have in mind (or a screenshot of some tooltip you like) rather than having to get into CSS yourself 😉

whyboris:
@cal2195 -- just let me know if you'd like me to re-open this PR 👌 (or re-open it yourself if you have the repository permissions 👍 )

Closing as I'm cleaning up before `2.0.0` release 🚀

Issue #131, opened 2019-01-23
Thumbnails don't show up if the path to the vha file has a space in it


cal2195:
Thumbnails also don't show up if:
- [ ] The path to the vha file has a space in it!

Reopening this issue as it's shown up again! 👍 

With mac, this is fixed with `encodeURI()`, but it seems (?) Windows can't handle `%20` in file names?

Maybe surrounding the path with quotes could help... investigation needed! 🔍

whyboris:
Since I have both a Windows and a Mac machine I'll look into it 👍

whyboris:
I'll look into it sometime this week and hopefully solve it without resorting to a hack: filling in the space character in the hub name with a `-` :trollface:

cal2195:
I identified a second issue which might be easy to fix at the same time - (silly) Windows treats the file string funny, and takes the `\` to mean an escape code! 😢 

This means that a path like `D:\Files\...` is taken as `D: iles\...`
(`\f` being form feed https://en.wikipedia.org/wiki/Page_break)

Encoding the string, or escaping everything should fix both issues! 👍

cal2195:
> I'll look into it sometime this week and hopefully solve it without resorting to a hack: filling in the space character in the hub name with a `-` :trollface:

This won't work, unless you plan on replacing all spaces in the entire path to the file as well... - :trollface:

cal2195:
Damn, this issue is going to be trickier than I thought - it also includes using a vha file created on one system, on another system... 😢 

Linux/Mac using `/` as a path delimiter, but Windows uses `\` - This means if you create on one, but then load on another, the saved paths aren't going to match `path.sep`! 👎 

Looking at [this](https://stackoverflow.com/a/31976060), it seems the best option would be to convert Windows `\` to `/` (it's used in the urls anyway), and then we can always assume folders are separated by `/` - I think this should work for opening files too.

Only problem is any libraries created on Windows will need to be rescanned - thumbnails should be safe! 👍 

Are you okay with this?

whyboris:
I think I understand your description of the solution you describe. Could you just clarify what you mean by:
> Only problem is any libraries created on Windows will need to be rescanned - thumbnails should be safe!
Meaning that after this change, any .vha file you create on Windows will work fine, but any old .vha file will not?

If so -- no problem at all -- version 2.0 is _all-breaking-changes_ (vha files, settings, etc - as if it's a separate app from before) 🙆‍♂️

[...]

PR #166, opened 2019-04-05
🎉 Show folders in views! 🎉

cal2195:
This _cool_ PR adds the option to show folders in the `thumbnails`, `files` and `clips` view! 🎉 

It basically replicates a standard file browser, which a lot of people will probably like - and the ability to browse to a folder, then flip the folder view off and continue like normal is _very_ cool! 😄 

It has some pretty snazzy features, which I'll leave you to try it to see what you think! 😉 

I also removed the `folder only` view, as I don't think it added anything with this option now - I can revert that commit if you think otherwise! 👍 

**Caveats**:
- `Folder Intersection` is treated as an exact prefix match if the first entry starts with `/`


cal2195:
Update: Works on Windows here for me, but only because I created the library on Linux - see #131

cal2195:
I haven't tested this on Windows yet, but this should now work with either an existing library after doing a meta regen in the settings, or on just a new library with this version! 👍 

I've also added support for sorting with folders enabled! 😄 🎉 Along with folder contents details!

cal2195:
Ready for review! 👍

cal2195:
These are going to have a few merge conflicts, so whenever you get a chance to check them out just let me know which you're good with and I'll address the conflicts and merge them! 👍

src/app/components/home/thumbnail/preview.component.ts
```diff
@@ -51,8 +54,30 @@ export class PreviewComponent implements OnInit {
...
+      this.numberOfThumbs = this.folderThumbPaths.length;
+    } else {
+      this.firstFilePath = 'file://' + this.folderPath + '/' + 'vha-' + this.hubName + '/thumbnails/' + this.video.hash + '.jpg';
+      this.fullFilePath =  'file://' + this.folderPath + '/' + 'vha-' + this.hubName + '/filmstrips/' + this.video.hash + '.jpg';
+      this.folderThumbPaths.push(this.firstFilePath);
+      this.numberOfThumbs = 1;
+    }
+  }
+
+  shuffle(a) {
```
whyboris: This method appears exactly the same in `clip.component` which is fine for now, but a few minor recommendations:
- include `JSDoc` to briefly describe what the method does
- add TypeScript to the input (so it's clear what kind of objects are allowed to be used as input)
- place the code for this method into some helper `service` and refer to the method from both places

All three can be accomplished later after the merge 👌

whyboris:
❤️ OMG you're amazing 🌟 

This is _so_ cool! 🙇

PR #170, opened 2019-04-08
Zoom by number of images per row

cal2195:
This PR implements #71, which is to zoom by an integer number of images/thumbnails per row! This allows the app to always use the full space available to it for displaying items! 👍 

This also fixes the issue from #95 in a nicer way! 🎉 

This PR is going to supersede #160, as I've integrated that change into this one! 😄

### DIFF
--- a/src/app/components/common/app-state.ts
+++ b/src/app/components/common/app-state.ts
@@ -21,7 +21,7 @@ export const allSupportedViews: SupportedView[] = [
   'showClips',
 ];
 
-export interface ImageHeights {
+export interface RowNumbers {
   thumbnailSheet: number;
   showThumbnails: number;
   showFilmstrip: number;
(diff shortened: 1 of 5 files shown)


cal2195:
This also improves the full view handling! 😄

whyboris:
I'm absolutely loving it! ❤️ 

My first reservation was about scaling images by not-nice fractions (not +50% but things like +37% could look ugly). But resizing to fit the width works just so beautifully!

I wonder if the app should update the screenshot size when the app is resized 🤔 -- unsure how often and when users resize the app (and what they would prefer to happen when they do). 

The size increase from 2 screenshots per row to 1 screenshot per row is really large; but I'm guessing there's little reason to want an in-between size 💭 ... only likely to matter in the `details` view 👌 

I think after this merges, we'll need to adjust this feature a bit for the _details_ view, to prevent the details view to expand to full-width (it then hides the details). 

I think the app is much better with this feature 👍 ❤️ 💯 

AMAZING work 🙇

PR #172, opened 2019-04-30
Update ngx-virtual-scroller

whyboris:
Update the `ngx-virtual-scroller` dependency and fix up a major scrolling issue we're experiencing currently in _master_:
- Thumbnail view will now add/remove elements as you scroll

There is still some jankiness when scrolling some views, resizing doesn't always seem to update the height for virtual scroller, so I'll create follow-up commits or other branches to fix those 👍

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "@ffmpeg-installer/ffmpeg": "1.0.17",
     "@ffprobe-installer/ffprobe": "1.0.9",
     "@ngx-translate/core": "11.0.1",
-    "ngx-virtual-scroller": "1.0.16"
+    "ngx-virtual-scroller": "2.0.7"
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "0.13.1",
--- a/src/app/components/home/details/details.component.scss
+++ b/src/app/components/home/details/details.component.scss
@@ -1,10 +1,10 @@
 .filmstrip-container {
   display: inline-block;
-  margin: 20px 20px 0;
+  min-height: 170px;
   overflow: hidden;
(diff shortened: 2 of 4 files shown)


cal2195:
This looks great! 👍  I'll check it out later! I was aware of this issue, glad you figured out a way to fix it! 🎉

whyboris:
I accidentally created the branch off the `button-cleanup` branch, but they are otherwise entirely independent (no conflicts). I'm happy to merge in both once you approve and I'll make new branches with further scroll improvements 👍 

Thank you for your recent feature additions -- version 2.0.0 is going to be so much better than I originally planned thanks to your work 🙇

whyboris:
ps -- some of the jumping while scrolling resulted because I was using `margin` rather than `padding`. It was the sensible way to describe the elements, but the virtual-scroll library has a minor bug: https://github.com/rintoj/ngx-virtual-scroller/issues/286#issuecomment-465838151 👌

whyboris:
Thanks for pointing that out! I was only working with about 30 videos in a hub, but I just tried with 600 -- I see the problem when resizing. But it seems to resume to working fine when you zoom in/out, then switch around to a different view (_full view_ perhaps), and then scrolling works fine again.

So I think there's a way to fix it up, perhaps with some method call to the scroller to recompute/refresh.

I think we definitely should update to the newest version of `ngx-scroller` to be future proof. The other changes are just `margin` -> `padding` (which created a bug that text overflows under thumbnails, but that's just some CSS to fix).

I also haven't tested with variable video lengths & full view yet -- so that could be an additional problem. But full view has similar trouble in _master_.

whyboris:
I'm merging this with full intention to fix up all the scroll/zoom bugs in subsequent PRs. It will be my top priority after merging in #174 👍

PR #174, opened 2019-05-02
Similar videos tab

### LINKED ISSUE #173: Similar videos tab
On _YouTube_, when you watch a video, there are _recommended videos_ in the sidebar.

This app can do the same very easily by reusing the already existing "show similar" functionality. 

There would be a button in the ribbon which would toggle a horizontal drawer along the bottom which would show 5 most-similar videos to the last video clicked.

Unsure how useful it would be to anyone, but it's easy-enough to implement. I'll add it 👍

whyboris:
Works 😎 

Closes #173 

Check it out 😉 

⚠️ this PR reorganizes the variables (line locations) into coherent sets and renames some of them 👍 

It's based off `virtual-scroller-2` branch -- so there's a chain of 3 dependent PRs now :trollface: I'll merge them in a sensible order once all 3 are approved 👍


whyboris:
The intention is the final feature will always show 5 videos. This will require adjustments to the current state as of last commit 🙆‍♂-- currently switching between views messes with the sizes of recommended videos (not desired behavior). 

I've not tested this feature for different widths of the app -- I think a good behavior would be to simply show as many as the width of the app can fit (with a pre-determined height of the tab) ... should be easy 👍

whyboris:
I should have commented in the PR this wasn't a solid finished feature 😅 

I think the best behavior would be to have a max height and show as many as fit with some minimum perhaps. Responsive behaviors take some care and I was just getting the feature draft out for quicker feedback.

There's some bug at the moment that titles don't show up in certain views, I'll investigate and debug.

I'm unsure what the "similar videos" would be of unless you're playing a particular video. There _is_ a right-click-on-video-show-similar feature in the app, but behaves differently a bit. Could you elaborate on what you had in mind?

whyboris:
Looks like the `padding` vs `margin` bug in _ngx-virtual-scroller_ has been fixed:
https://github.com/rintoj/ngx-virtual-scroller/issues/286#issuecomment-489375914
I'll pull in the new version and update the css 👍

whyboris:
😎 turns out all we needed was `invalidateAllCachedMeasurements()` and smooth scrolling is back after zooming in/out 😎 

🙏 🙇 thank you to `ngx-virtual-scroller` 💯 🎖 🏅 🥇

PR #176, opened 2019-05-08
Update Electron to 5.0.1

whyboris:
And update as many dependencies as I can without breaking 🙆‍♂ 

It's unclear if we need `node-sass` -- we were fine beforehand, but that could have been because it was implicitly included with _Angular-CLI_ ... but as I was jumping around to _Node 12_ it seemed to have messed things up. There's little harm in having an extra _devDependency_ 👌


whyboris:
Update all the versions of dependencies (other than Angular) that can be updated without breaking ✅ 

We'll update to Angular 8 once it comes out 👍 -- it might make it easier to update to Electron 5 👍 and might solve the problems with updating `zone.js` and `core-js` 😅

whyboris:
Confirming I'm able to run this branch with _Node 12_ :tada:

I recommend using `nvm` to switch versions (rather than install a new node version globally -- just in case). 

Might require running `npm rebuild node-sass` after switching to Node 12 👌

whyboris:
🎉 Was able to update to `Electron 5.0.1` thanks to this:

https://stackoverflow.com/questions/55093700/electron-5-0-0-uncaught-referenceerror-require-is-not-defined

cal2195:
This looks great! I'll try it out now! 👍

PR #181, opened 2019-05-09
Actually quit the app

cal2195:
This PR just removes a function that prevented the app from ever actually closing when you quit! 👎 

I think it was added to allow `cmd+q` to process, but this is already in the keycode handling, and it doesn't seem to be required here anymore! 😄

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -154,10 +154,6 @@ try {
     // }
   });
 
-  app.on('before-quit', (event) => {
-    event.preventDefault();
-  });
-
   app.on('activate', () => {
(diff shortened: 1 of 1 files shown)


whyboris:
Thank you for keeping vigilant and cleaning up old code 👀 

I'll have to build the app to test this code change. Might take a week or so (just need to find some time) - thank you ahead of time for your patience. 

Cheers 🙌

whyboris:
I see this was part of #79 -- hope the problem of not saving the settings doesn't resurface 😅

PR #182, opened 2019-05-13
bump electron-builder version

whyboris:
🚀 

I compiled the app `npm run electron:mac` and it worked 👌 

Glad to see bumping version doesn't break things (as expected, though there _are_ times when it does 😅 )

Merging -- no need to review 👌


cal2195:
🚀

PR #189, opened 2019-06-09
Clip fixes

### LINKED ISSUE #185: Clip fixes
I'd like to have an option to disable sound on the _clip_ view
Also have a short random timeout before a clip starts playing (when all clips are playing) so that the app doesn't try to access 20 files all at the same time 😅

whyboris:
Closes #185 

- Mute video settings button
- random delay in starting playback (reduces synchronized frame switching which is visually unpleasant)
- button appear/disappear animation (clicking on the eye icon animates appearing ribbon icon)
- simpler button descriptions
- _code change_ sort settings buttons object by key alphabetically 

⚠️ no special icon was created yet so the button is really crappy 😅


whyboris:
Hey @cal2195 -- no need for review; though feel free to comment if you think something is off.

- The button description text isn't final (just better than before)
- audio isn't disabled everywhere when the _mute_ button is toggled, I'll fix that up later 👌

src/app/components/home/clip/clip.component.ts
```diff
@@ -28,6 +28,7 @@ export class ClipComponent implements OnInit {
   @Input() hubName: string;
   @Input() imgHeight: number;
   @Input() largerFont: boolean;
+  @Input() mutedNow: boolean;
```
cal2195: Would `forceMute` be more appropriate?

src/app/components/home/clip/clip.component.html
```diff
@@ -145,6 +145,7 @@
     [ngStyle]="{ 'height': imgHeight + 'px' }"
     (click)="videoClick.emit($event)"
     class="screencap"
+    [volume]="shouldMute()"
     onmouseover="this.muted = false;"
```
cal2195: Might make more sense to have
```
if (!this.forceMute) { this.muted = false };
...
```

Thoughts? 😄

whyboris: What about just bringing the `shouldMute()` method right into the template?
```
[volume]="forceMute ? 0 : 1"
```
If I remember right I had trouble with muting -- since hovering over the element toggles the `muted` attribute, hence why I resorted to the hack with the `volume` property 😅

cal2195: Well I was more thinking it might be better to disable the audio on hover behaviour, rather than also using the `volume` property? There's no point unmuting it if the volume is 0 anyway, but maybe your approach has less edge cases... 🤔

PR #192, opened 2019-06-18
version bumps

whyboris:
Nothing major; the biggest two are:
- electron-builder
- electron
But both are just minor version bumps 👌 

`npm run electron:mac` still works -- 🙆‍♂ 

Current Mac _dmg_ file size: 104MB


cal2195:
🎉

PR #193, opened 2019-06-29
Add pt_br translation

Lakshamana:
Please review


whyboris:
Before version 2 is released I might change some of the language to make the buttons / settings descriptions more clear. I'll then re-run the script to update the text and open a PR tagging you so you could glance it over (if you find the time).

Thank you again 👍

Issue #194, opened 2019-06-30
Propose of improvement of the translation addition workflow


Lakshamana:
Although it is easy to add new translations (because of translate.sh), the process of adding new translations is still manual (manual edition in some files) and then it could lead to mistakes in the time to add new languages. In order to minimize those mistakes and turn contributors more confident to this process, I propose here an automatic way:
Just run (for instance):
`./translate.sh Deutsch de`
and then it's done.
I was working on it and will submit a PR.

whyboris:
Thank you again for your work @Lakshamana 🙇 

Closed with #195

PR #195, opened 2019-06-30
Add translation process automation

Lakshamana:
Refers to: #194.
Please, review.

### DIFF
--- a/src/app/i18n/translate.sh
+++ b/src/app/i18n/translate.sh
@@ -49,13 +49,15 @@ get_os() {
   return 0
 }
 
+# check OS
+get_os
+
 # check if NEW_LANG_CODE is in app-state.ts
 _list1=`cat ../components/common/app-state.ts | grep 'export type SupportedLanguage' | cut -d'=' -f 2 | sed "s/^ '//" | sed "s/';$//" | sed "s/' | '/ /g"`
 in_list $NEW_LANG_CODE "$_list1"
 if [ $result = "true" ]; then
   echo $NEW_LANG_CODE 'already in app-state.ts'
 else
-  get_os
   sed -i $SED_HACK "/export type SupportedLanguage/ s/;$/ | '$NEW_LANG_CODE';/" ../components/common/app-state.ts
   echo 'added' $NEW_LANG_CODE 'to app-state.ts...'
 fi
--- a/src/app/i18n/translate.sh
+++ b/src/app/i18n/translate.sh
@@ -77,7 +77,7 @@ else
     echo $NEW_LANG_CODE ' case already added'
   else
     get_os
-    sed -i $SED_HACK "/this.appState.language = '$last'/ s/$/\n        break;\n      case '$NEW_LANG_CODE':\n        this.translate.use('$NEW_LANG_CODE');\n        this.translate.setTranslation('$NEW_LANG_CODE', $NEW_LANG_NAME );\n        this.appState.language = '$NEW_LANG_CODE';/" ../components/home/home.component.ts  sed -i "/i18n\/$last/ s/$/\nimport \{ $NEW_LANG_NAME \} from '..\/..\/i18n\/$NEW_LANG_CODE';/" ../components/home/home.component.ts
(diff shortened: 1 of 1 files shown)


whyboris:
Thank you for this awesome contribution -- I'll test it out within a day or two and likely merge as is 👍

Lakshamana:
@whyboris, I'm using Manjaro Linux. I'll start to test your OS X solution for sed. For sake of portability, it could be handy to add a function to get system's label (bash's `uname`* command). Thanks for reviewing!

whyboris:
I already tried my solution on a Mac -- it worked :)
I think it's just worth seeing if including the `""` still works on Linus -- if it does, we may as well merge it in 👌 

I always feel uneasy whenever I request any edits to a PR -- a person (you in this case) have already done some awesome work, and to request more feels like in poor taste 😅 

I'm happy to merge the feature as is too -- no pressure to do more work 🙇 🙇 🙇

Lakshamana:
Hey, @whyboris, I just added a function to check whether it is running on a Mac or Linux (and could be added more OS cases). If it's a Linux, use Linux's `sed` otherwise it will use Mac's one

whyboris:
I'm unsure what the problem with _Mac_ is -- but the script generates new files on my machine:
```
        "../components/common/app-state.ts\"\""
        "../components/home/home.component.html\"\""
        "../components/home/home.component.ts\"\""
```
Note the `\"\""` at the end. The files look like what we want, but the stuff at the end is likely some other OS X bug 😞 

ps - the OS detection code is in the first block, but the other `sed` commands would need the same treatment. Perhaps this would be easier? Use your code somewhere above to define a new variable:
```
get_os
if [ $result = 'linux' ]; then
  SED_HACK=''
elif [ $result = 'osx' ]; then
  SED_HACK='""'
else
  echo 'System not supported'
  exit 1
fi
```
Now insert the `SED_HACK` into the command, e.g.
```
sed -i $SED_HACK "/export type SupportedLanguage/ s/;$/ | '$NEW_LANG_CODE';/" ../components/common/app-state.ts
```
So on _Linux_ this hack would insert an empty string, not causing trouble, while on _Mac_ we'll have the desired `""` 🤔

cal2195:
Automation++ 😄 

I wonder if this would be easier if we had a separate file for the LANG_CODES, and dynamically populated them at runtime in each location, so we only need to edit one file?

Also, this won't run on Windows by default, which is why I used node for the original script. not sure if that's an issue or not? 🤔 

Great work @Lakshamana 👍

Lakshamana:
> ```
> get_os
> if [ $result = 'linux' ]; then
>   SED_HACK=''
> elif [ $result = 'osx' ]; then
>   SED_HACK='""'
> else
>   echo 'System not supported'
>   exit 1
> fi
> ```
> 
> Now insert the `SED_HACK` into the command, e.g.
> 
> ```
> sed -i $SED_HACK "/export type SupportedLanguage/ s/;$/ | '$NEW_LANG_CODE';/" ../components/common/app-state.ts
> ```
@whyboris, yeah, that approach would be easier. I was thinking in such a way to do this but didn't figure out a way to do this properly.

Lakshamana:
Thanks, @cal2195. If you prefer I could migrate my code to node. However, Windows 10 can run bash, though it is a beta feature (and it's not a native windows application). Since I don't use windows at all, I don't know if all bash commands are available for it to use.

cal2195:
@Lakshamana If this is working, I'd say merge as is for now! Just something we could do in the future, especially if we wanted to add more automation to this/other tasks! 😄 

WSL is certainly good, and should basically appear as a Linux system, but if you're doing VHA dev on Windows, you already have node, so it's one less thing to get installed! 😅

Lakshamana:
@cal2195, totally agree on dev dependencies for improving dev experience. This works in my machine. So it's ready to merge (for now, I'm just waiting for tests in OS X).

whyboris:
This worked beautifully on my _Windows_ machine ✅ 
On _Mac_ it still created 3 new files rather than update the old ones (see https://github.com/whyboris/Video-Hub-App/pull/195#issuecomment-509895401).
I'll merge as is -- it's probably just a few characters away from perfect and I'll get it to work and patch it in the near future 🤝 

Thank you again for your work @Lakshamana and thank you for joining in @cal2195 🙇

PR #196, opened 2019-07-02
Stylelint for all the SCSS files

whyboris:
[Stylelint](https://stylelint.io/) is awesome for keeping all the CSS properties alphabetized for easier reference 👍 

There are some `do not alphabetize` comments in the code I'll need to double-check and make sure nothing broke visually 😅

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,9 @@
     "electron:local": "npm run build:prod && electron .",
     "electron:linux": "npm run build:prod && electron-builder build --linux",
     "electron:windows": "npm run build:prod && electron-builder build --windows",
-    "electron:mac": "npm run build:prod && electron-builder build --mac"
+    "electron:mac": "npm run build:prod && electron-builder build --mac",
+    "lint:scss": "stylelint \"./src/**/*.scss\"",
+    "lint:scss:fix": "stylelint \"./src/**/*.scss\" --fix"
   },
   "dependencies": {
     "@ffmpeg-installer/ffmpeg": "1.0.18",
@@ -63,6 +65,10 @@
     "node-sass": "4.12.0",
     "npm-run-all": "4.1.5",
     "rxjs": "6.5.2",
+    "stylelint": "10.1.0",
+    "stylelint-config-recommended-scss": "3.3.0",
+    "stylelint-order": "3.0.0",
+    "stylelint-scss": "3.8.0",
     "ts-node": "8.3.0",
     "tslint": "5.17.0",
     "typescript": "3.4.5",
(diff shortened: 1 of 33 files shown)


src/app/components/home/layout.scss
```diff
@@ -151,14 +151,14 @@ $sidebar-width: 190px;
   box-sizing: border-box;
   height: 100vh;
   max-width: $sidebar-width !important;
+  // don't alphabetize for some reason
```
whyboris: I'll need to check this manually 🙆‍♂

src/app/components/home/buttons.scss
```diff
@@ -48,13 +48,13 @@ button {
...
   fill: #729aff;
-  transition: $button-toggle-length;
+  transition-duration: $button-toggle-length;
 }
 
 .defaultSettingsButton {
-  border: 1px solid $gray-30; // don't alphabetize this one
 
   background: $gray-15;
+  border: 1px solid $gray-30; // don't alphabetize this one
```
whyboris: Check manually

src/app/components/home/layout.scss
whyboris: Resolved ✅

src/app/components/home/buttons.scss
whyboris: Resolved ✅

package.json
```diff
@@ -34,7 +34,9 @@
     "electron:local": "npm run build:prod && electron .",
     "electron:linux": "npm run build:prod && electron-builder build --linux",
     "electron:windows": "npm run build:prod && electron-builder build --windows",
-    "electron:mac": "npm run build:prod && electron-builder build --mac"
+    "electron:mac": "npm run build:prod && electron-builder build --mac",
+    "lint:scss": "stylelint \"./src/**/*.scss\"",
```
cal2195: Maybe we should include this in one of the building commands above? 🤔

whyboris: Would be good to automate the lint sometime, but I don't think it would be a good idea during the build.

I see the lint as a general house-cleaning procedure to keep our SCSS organized (more-readable / easier to search through). It's not important for the built app.

The biggest problem is that since the lint alphabetizes things, some properties might override the behavior of others above, resulting in an incorrect display of some elements. I'd rather have the build come out as-close to the dev environment as possible 😅

cal2195: To be clear, I meant `lint:scss` just to display the warnings (I assume that's what it does), rather than the `lint:scss:fix` version to actually change the code! 👍 

That way you're more likely to notice, oh, I should change that, rather than remembering to run the linter! 😅

PR #197, opened 2019-07-04
Video Hub App 2

### LINKED ISSUE #168: Rename .vha to .vha2
In the same spirit as #158, should we also bump the file extension of version 2.0.0 compatible library files as old .vha files will not be compatible?

Maybe `.vha2`? 🤔 💡

whyboris:
Sike! 
:trollface: 

It's just a name change. Now the app 
- will install as `Video Hub App 2`
- only open and save `.vha2` files
- have its own `settings.json` 

This will allow users to install version 2 while still being able to use version 1 and all the hubs they've created. 🎉 

Closes #168

### DIFF
--- a/src/app/components/home/home.component.ts
+++ b/src/app/components/home/home.component.ts
@@ -578,7 +578,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
       if (ev.dataTransfer.files.length > 0) {
         const fullPath = ev.dataTransfer.files[0].path;
         ev.preventDefault();
-        if (fullPath.slice(-4) === '.vha') {
+        if (fullPath.slice(-5) === '.vha2') {
           this.electronService.ipcRenderer.send(
             'load-this-vha-file', ev.dataTransfer.files[0].path, this.saveVhaIfNeeded()
           );
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -1,5 +1,5 @@
 {
-  "productName": "Video Hub App",
+  "productName": "Video Hub App 2",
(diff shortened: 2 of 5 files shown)


whyboris:
Works well on Windows too 👌 _merging_

src/app/components/home/home.component.ts
```diff
@@ -578,7 +578,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
       if (ev.dataTransfer.files.length > 0) {
         const fullPath = ev.dataTransfer.files[0].path;
         ev.preventDefault();
-        if (fullPath.slice(-4) === '.vha') {
+        if (fullPath.slice(-5) === '.vha2') {
```
cal2195: I feel like we should have some global `FILE_EXTENSION='.vha2'` for everywhere that uses the extension, and then also should this be `fullPath.endsWith(FILE_EXTENSION)`?

whyboris: Thanks for the `.endsWith()` method -- I've not come across it before 😰 it's great to know it's available 😍 

Having a global would make search-and-replace unnecessary if we ever upgrade to `.vha3` :trollface: 

Unsure what other benefit we'll get by having a constant rather than a string ... omitting spelling mistakes is the only one that comes to mind now 🤔

cal2195: Well, I count 6 `.vha2` occurrences in this change - I'd do it for anything over 2! 😄 

Maybe a better name like `globals.ext` would be prettier! 👍

PR #199, opened 2019-07-04
All extraction code in its own file and uses Promise

### LINKED ISSUE #198: Option to (not) generate clips
Include an option in the _wizard_ for user to (not) generate clips.

Currently clips are automatically generated. I think this option will be disabled by default (unchecked check box), but will be easy to enable during the wizard.

Since someone might not initially want them, but later changes their mind, I'll need to add an extra option in the _settings_ to import clips. So there will be yet another boolean in the `FinalObject` 👌 

Finishes what was started with #17

whyboris:
Huge thank you to @cal2195 for his work on the extraction code. Most of the code in the PR is just moving it around, refactoring a bit, and providing a UI interface to the variable he exposed 🙇 

A refactor to isolate code into its own file and to use `Promise` for the extraction procedure.

The code should be simpler to follow:
- There is a now a dedicated file for all screenshot / clip extraction: `main-extract.ts`
- All helper functions now return a `Promise` 
- The extraction loop happens in a method `extractFromTheseFiles` which cleanly runs through each step in sequence.

Additionally:
- User can choose not to extract video clips
- User can choose the number of `clipSnippets` (how many sections of original to put into the preview clip)
- User can choose the length of each clip snippet 😎 

Closes #198 

To do in another PR:
- fix a bug where if you include a .jpg file renamed as .mp4 and extract, the node console keep scrolling with empty new lines forever 😅 
- check whether re-scan works as desired
- check whether re-scan avoids generating clips when `clipSnippets = 0`


whyboris:
Works on Windows 👌 _merging_

main-extract.ts
```diff
@@ -0,0 +1,488 @@
...
+        theFinalArray[currentElement].partialPath,
+        theFinalArray[currentElement].fileName));
+
+      const duration: number = theFinalArray[currentElement].duration;
+      const fileHash: string = theFinalArray[currentElement].hash;
+      const numOfScreens: number = theFinalArray[currentElement].screens;
+
+      checkIfInputFileExists(pathToVideo)
+
+        .then(content => {
```
cal2195: Now this is cool!! ❄️

whyboris: Teamwork 🤝

PR #202, opened 2019-07-08
Version bump

whyboris:
Just keeping up to date 🙆‍♂


whyboris:
Tested on _Windows_ and _Chrome Dev Tools_ throws some kind of error into Node 
```
Cannot read pseudoType
```
Not a big deal -- everything runs fine, but a bit annoying. I'll delay merging for a bit 🙆‍♂

whyboris:
Yeah -- I suspected that the problem might be with a new version of Chrome included in the new _Electron_ (though I'm just guessing -- unsure if that's what the update includes).

I think the next major version bump in Electron is coming, so I might just upgrade then 👌 

It's a bit silly for me to be chasing all the new updates -- the app doesn't interact with the internet so I don't need any security/vulnerability patches ... and there isn't some new Electron API I'm hoping to use ... just doing some _code gardening_ 🏡

cal2195:
Haha, no not at all! 😄 I'd ***way*** prefer doing many small updates with little changes, than need to work out how to jump 3 major updates to get that new critical feature you want!

Updates are good! 👍

whyboris:
Checked today -- works without problems on _Mac_
Then I checked Windows again -- no error I saw before! 🎉 
Also ran `npm audit fix` 💯 
Merging ✅

PR #203, opened 2019-07-10
Handle folders with spaces

### LINKED ISSUE #131: Thumbnails don't show up if the path to the vha file has a space in it

whyboris:
Bugfix for Windows:
- handle opening `vha` files from paths that include folders with spaces 😱 

Huge thank you to @cal2195 for diagnosing the problem and finding a solution (see #131)

Closes #131 

⚠️ Seems to create a new bug -- when creating a new _hub_ the previews are black until you exit the app and re-open it. I think it should be a quick fix -- though I might do it later (it's late tonight 😴 )

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -302,7 +302,7 @@ function openThisDamnFile(pathToVhaFile: string) {
         'finalObjectReturning',
         lastSavedFinalObject,
         pathToVhaFile,
-        globals.selectedOutputFolder + path.sep,   // app needs the trailing slash (at least for now)
+        getHtmlPath(globals.selectedOutputFolder),
         changedRootFolder,
         rootFolderLive
       );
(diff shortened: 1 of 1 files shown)


main.ts
```diff
@@ -302,14 +302,36 @@ function openThisDamnFile(pathToVhaFile: string) {
...
+ * e.g. `C:\Some folder` becomes `C:/Some%20folder`
+ * @param anyOsPath
+ */
+function getHtmlPath(anyOsPath: string): string {
+      // Windows was misbehaving
+      // so we normalize the path (just in case) and replace all `\` with `/` in this instance
+      // because at this point Electron will be showing images following the path provided
+      // with the `file:///` protocol -- seems to work
+      const normalizedPath: string = path.normalize(anyOsPath);
+      const forwardSlashes: string = normalizedPath.replace(/\\/g, '/');
```
cal2195: This should already be done everywhere else as of c37341c? Are you still having issues?

main.ts
```diff
@@ -302,14 +302,36 @@ function openThisDamnFile(pathToVhaFile: string) {
...
+      return forwardSlashes.replace(/ /g, '%20');
```
cal2195: I thought this didn't work on Windows - otherwise we could just use `encodeURI()`?

whyboris: The biggest problem is that the URL string is part of a conditional inside _ngStyle_ in a template:
```
[ngStyle]="{ 'background-image': somethingTrue ? oneURL : anotherURL }"
```
I guess it could be formatted beforehand in the template with `encodeURI()` too (I'll check).

Today again it's too late to investigate 😴 -- but I'll look into all places we're dealing with slashes (thank you for the https://github.com/whyboris/Video-Hub-App/commit/c37341c100b5514fe161d32e48e40534307c7829 ref) and see if I can consolidate things 😅

main.ts
whyboris: Before this change, the `globals.selectedOutputFolder` was getting garbled before it got rendered through the template.

This should fix it once and for all 😅

whyboris:
I'll be double-checking everything thoroughly on _Mac_ and _Windows_ before release. I'll be checking folders that have capital letters, spaces in them, etc. I'm currently eager to close as many issues as I can to get version 2 released 🚀 

Thanks for your constant vigilance 🙇

PR #208, opened 2019-07-13
Fix and also disable checkAllScreensExist

whyboris:
Turns out the `checkAllScreensExist` method used the `ffmpeg` command without `spawn`, thus making the app try to use the system `ffmpeg` 😱 

This worked until I reinstalled my laptop yesterday to start working with a clean slate and noticed this method was blocking extraction :trollface: 

Currently enabling the method runs well, but the `console.log` (line 87) keeps emitting numbers all after the extraction is finished.

Sadly I'm not sure what the ffmpeg arguments do:
```
-v warning -ss ' + time + ' -t 1 -i "' + pathToVideo + '" -map V -f null -
```
@cal2195 -- let me know if you're interested in debugging the arguments / method after this code merges 🙇

### DIFF
--- a/main-extract.ts
+++ b/main-extract.ts
@@ -77,33 +77,54 @@ const checkAllScreensExist = (
     const totalCount = numberOfScreenshots;
     const step: number = duration / (totalCount + 1);
 
+    // check for complete file
     const check = (current) => {
       if (current === totalCount) {
         resolve(true);
       }
-      // check for complete file
+
+      // !!!!!!
+      // Notice this keeps running even after screenshots successfully extracted!
+      // !!!!!!
+      console.log('checking ' + current);
+
       const time = (current + 1) * step; // +1 so we don't pick the 0th frame
-      const checkCommand = 'ffmpeg -v warning -ss ' + time + ' -t 1 -i "' + pathToVideo + '" -map V -f null -';
       const corruptRegex = /Output file is empty, nothing was encoded/g;
-      exec(checkCommand, (err, data, stderr) => {
-        console.log(data);
-        console.log(stderr);
-        if (err) {
-          // skip this file
-          console.log(pathToVideo + ' was corrupt, skipping!');
-          resolve(false);
-        } else {
-          if (data.match(corruptRegex) || stderr.match(corruptRegex)) {
+
+      const args = [
+        '-v', 'warning', '-ss', time, '-t', '1', '-i', pathToVideo, '-map', 'V', '-f', 'null', '-',
+      ];
+      // console.log('extracting clip frame 1');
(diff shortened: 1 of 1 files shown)


whyboris:
ps - I remember the motivation behind this method was to detect whether a file is corrupt (at least in the spots where we were going to extract screenshots) and if so, to _not_ extract screenshots (to prevent a nasty error of some kind).

I suspect this whole method slows down the extraction process by quite a bit ... so it would be excellent if we just caught the errors at the time of extracting the screenshots, rather than independently before 🤔 

@cal2195 -- could you remind me what the error was; could it be 🤞 that now with `Promise` it will not error as hard and will work better? Perhaps you could try to replicate the error you used to get?

🙇 thanks for your time!

main-extract.ts
```diff
@@ -77,33 +77,54 @@ const checkAllScreensExist = (
...
-          console.log(pathToVideo + ' was corrupt, skipping!');
-          resolve(false);
-        } else {
-          if (data.match(corruptRegex) || stderr.match(corruptRegex)) {
+
+      const args = [
+        '-v', 'warning', '-ss', time, '-t', '1', '-i', pathToVideo, '-map', 'V', '-f', 'null', '-',
+      ];
+      // console.log('extracting clip frame 1');
+      const ffmpeg_process = spawn(ffmpegPath, args);
```
cal2195: Good catch here! 👍  `exec` would also work as before, but should've had `ffmpegPath` not `ffmpeg` hardcoded!

main-extract.ts
```diff
@@ -77,33 +77,54 @@ const checkAllScreensExist = (
...
+      // Note from past Cal to future Cal:
+      // ALWAYS READ THE DATA, EVEN IF YOU DO NOTHING WITH IT
+      ffmpeg_process.stdout.on('data', function (data) {
```
cal2195: I can't remember, but I might have chosen `exec` for a reason... let me think about it!

cal2195: Yeah, I think `exec` might be better - `spawn` uses a stream, so will keep running a bit after we decide the file is corrupt, while `exec` will give us the whole buffer once the command completes.

That way we know the command has finished, and we can start the next one.

cal2195:
Re: `-v warning -ss ' + time + ' -t 1 -i "' + pathToVideo + '" -map V -f null -`

`-v warning` - output all warnings (so we can match on them)
`-ss time` - jump into the clip at a certain time
`-t 1` - grab 1 second of video
`-i video` - path to the video
`-map V` - map the first/best video stream (can't remember off the top of my head)
`-f null` - send the resulting conversion to `/dev/null` (throw it away)
`-` - send (something) to stdout

PR #209, opened 2019-07-13
mute works across all clip view variations

### LINKED ISSUE #115: Clip gallery fixes
Two minor possible improvements to the video gallery (with live video previews) 

- Stop playing videos when the app is minimized
- Add a mute button (on by default) so hovering over a video does not play the audio

While we could simply not extract audio (it ends up choppy since the clips are short and jump between scenes), but since we already have the audio extracted (a freebie!) we may as well just have a mute button that is on by default. Thoughts? 😸

whyboris:
Mute toggle now mutes all sounds from every clip preview type
- regular clip
- regular clip (autoplay enabled)
- folder of clips
- folder of clips (autoplay enabled)

If _autoplay_ is enabled, the videos stop playing if windows loses focus (happens whenever minimizing the app) and restarts playing when window regains focus. 
- as a side-effect, at the moment, clicking into another window stops the auto-play (could be good behavior), but it's implemented that all the previews become black (rather than going back to preview images). I can fix that later.

Closes #115

### DIFF
--- a/src/app/components/home/clip/clip.component.html
+++ b/src/app/components/home/clip/clip.component.html
@@ -20,6 +20,7 @@
     [src]="'file://' + folderPath + '/' + folderThumbPaths[0]"
     [poster]="sanitizer.bypassSecurityTrustUrl('file://' + folderPath + '/' + folderPosterPaths[0])"
     class="mini-video-thumb"
+    [volume]="forceMute ? 0 : 1"
     onmouseover="this.play().catch((err) => {this.load()})"
     onmouseout="this.pause()"
     loop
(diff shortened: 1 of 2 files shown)


src/app/components/home/clip/clip.component.html
```diff
@@ -133,13 +141,14 @@
...
     class="screencap"
+    [volume]="forceMute ? 0 : 1"
     onmouseover="this.play().catch((err) => {this.load()})"
     onmouseout="this.pause()"
     loop
   ></video>
 
   <video
-    *ngIf="noError && autoplay"
+    *ngIf="noError && autoplay && appInFocus"
```
cal2195: Will this cause the videos to reload again from disk?

whyboris: Oh! Good catch! I'm unsure 😅 

Even if that happens, it's probably better than running at 100% CPU to render everything in the background 😕 ... the _autoplay everything_ is a really CPU-hungry feature 😈

PR #210, opened 2019-07-13
thumbnail autoadvance & setting toggle & refactor

### LINKED ISSUE #39: Auto play 10 screenshots on hover option

whyboris:
Setting toggle to auto-advance thumbnails (so you don't have to scrub)
Small refactor on the thumbnail to optimize computation & remove `hostiListener` 👍 

Closes #39


cal2195:
Can't wait to try this out! 👍

PR #212, opened 2019-07-13
Duration filter improvement

### LINKED ISSUE #132: Duration filter issues
Some small issues with the current duration filter:

- [ ] It's possible to move less than min and greater than max!
- [ ] It's quite hard to accurately choose a time - maybe less divisions would be better?
- [ ] Consider using a logarithmic scale (do you really care about seconds above 30 minutes?)

whyboris:
Duration filter now spans from lowest to the `durationOutlierCutoff`, advancing the slider to the right-most position shows "every" (not i18n translated yet 😓 -- might not be in 2.0.0 since it's not straightforward and I'd rather fix other things first).

The cutoff is computed as 3 * IQR (Inter Quartile Range). This makes for a better duration filter experience, at least in my case 😅 @cal2195 -- if you have time, please let me know how it looks in your library 👍 

As a bonus, this PR fixes the CSS on the _star filter_ 💯 (was very easy, though it created a minor bug where `N/A` column doesn't show up any more sometimes 🤷‍♂ still way better than before 👍 

Closes #132

### DIFF
--- a/src/app/components/home/home.component.ts
+++ b/src/app/components/home/home.component.ts
@@ -257,7 +257,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
   manualTagFilterString: string = '';
   manualTagShowFrequency: boolean = true;
 
-  longest: number = 0; // longest duration of all the video files -- only used in the Duration Filter
+  durationOutlierCutoff: number = 0; // for the duration filter to cut off outliers
 
   // ========================================================================
   // Please add new variables below if they don't fit into any other section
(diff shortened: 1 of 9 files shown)


src/app/components/home/home.component.ts
```diff
@@ -1729,4 +1716,44 @@ export class HomeComponent implements OnInit, AfterViewInit {
...
+   * defined unconventionally as "anything beyond the 3rd quartile + 3 * IQR (the inter-quartile range)"
+   * @param someArray
+   */
+  getOutlierCutoff(someArray: number[]): number {
+    const values = someArray.slice();
+    values.sort((a, b) => { return a - b; });
+
+    const q1 = values[Math.floor((values.length / 4))];
+    const q3 = values[Math.ceil((values.length * (3 / 4)))];
+    const iqr = q3 - q1;
```
cal2195: ❤️

Issue #205, opened 2019-07-10
Re-scan options


whyboris:
### Catch-all of thoughts about re-scan options

⚠️ Check all re-scan code (since the extraction code was refactored recently).

⚠️ Make sure re-scan can happen quickly.

It's possible that users may not want to extract _clips_ into their _hub_, but later might change their mind. It would be nice to have a setting that they can edit _after_ the initial scan. This might reside in the _statistics_ at least for now. 

ps - consider adding into the _statistics_ the size of the hub on hard disk.

Perhaps it should be possible to change the number of screenshots to take per video after the initial scan. This too might reside in the _statistics_ tab in the settings menu.

_Maybe(?)_ include a toggle to re-scan the hub on-open (unsure whether this should be a global setting or a per-vha file (saved inside the `vha` file).

A user once requested a right-click 're-scan' though it's not clear whether that's needed once the general re-scan functionality is solid.

cal2195:
> Perhaps it should be possible to change the number of screenshots to take per video after the initial scan.

If so, it would be nice to have the _option_ to regenerate all current clips to reflect the new setting! 👍

whyboris:
#215 is a start of this saga 😅 

It first moves all the relevant methods inside `main.ts` to a single section and moves all related methods from `main-support.ts` into its own file `man-rescan.ts` 👍 

From there it will be easier to see what methods depend on what other methods, and whether there's any duplication of code (trying to keep things _DRY_).

After merging I'll take stock of what methods we already have and what functionality we are missing 🙌

whyboris:
We currently have four re-scan buttons to click on:

https://github.com/whyboris/Video-Hub-App/blob/master/main.ts#L743

1. import-new-files
 - re-scan for all changes (addition, renaming, deleting)
2. rescan-current-directory
 - quick re-scan for only new videos
3. regenerate-library
 - regenerate all metadata except for thumbnails
4. verify-thumbnails
 - import any missing thumbnails

I'm a little tempted to comment out (3) for version 2 release -- it seems not needed unless the app starts tracking new metadata that was never extracted previously? @cal2195 -- no rush, but when you have a chance, could you comment on whether I understood (3) correctly, and perhaps share a use case other than the one I mentioned?

cal2195:
[...]

PR #215, opened 2019-07-14
Rescan refactor

whyboris:
Actual code change:
- respect user's preference for whether to extract _clips_ 👍 

Simply moving methods around:
- ⚠️ refactor/reorganize all _rescan_ code into its own file 😅 
- group _rescan_ methods into coherent sections in other files 😂

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ main.js
 main-extract.js
 main-filenames.js
 main-globals.js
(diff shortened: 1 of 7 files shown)


whyboris:
Merging so I can start on the next step of #205

cal2195:
Looks good to me! 👍

PR #217, opened 2019-07-14
Settings cleanup

whyboris:
- Organized settings
- Updated a lot of the text for consistency
- Removed `randomImage` button
- Made `shuffleGalleryNow` work again  
  - as a bonus - updates the sidebar dropdown if visible 😎 
- Hid `randomizeGallery` 
  - it was an option to always shuffle the results after every search
  - don't think it will be back in 2.0.0, if ever 🤷‍♂ open to comments

Oh we're getting closer and closer to `2.0.0` 🎉


cal2195:
> `it was an option to always shuffle the results after every search`
Is this not possible anymore? 😢

whyboris:
Now there is the `Sort order` dropdown -- and you can select `Random` there -- and every time anything happens, the results will be randomized 🎉

PR #221, opened 2019-07-21
normalize path before open

whyboris:
Was erroring out on _Windows_ when trying to open a path - because the file included a mix of `\` and `/` 😞 

Now fixed 👌


cal2195:
👍

PR #225, opened 2019-07-25
better stats & re-scan improvements

### LINKED ISSUE #216: Change Detection does not run often enough
Sometime before (perhaps before the _Angular 8_ upgrade), change detection ran often enough that the app was responsive.

Now, there are instances where change detection doesn't run like before, e.g:
- Select the _input_ folder in the _new hub wizard_ and until you click, nothing changes
- Start the app, get gray screen until you click anywhere, then it loads
- etc

Perhaps it's just a setting in Angular config somewhere to make change detection occur more often, or I'll just manually do `cd.detectChanges()` in appropriate places 🤷‍♂ 

On my to-do list ✅

whyboris:
- Stats now show screenshot settings 💇‍♂ 
- Cleaner interface for the `extractFromTheseFiles` method
- Closes #216 -- should run change detection often enough 😅 

Todo:
- allow user to enable importing of clips ... maybe 😅 
- pass in screenshot settings to rescan method (if user allowed to change the settings)

Continuing work on #205 🗻


cal2195:
Great work! 👍

PR #226, opened 2019-07-27
no clips extracted message when appropriate

### LINKED ISSUE #204: Hide clip view when no clips available
Hide the button to switch to the `clip` view if the hub does not contain clips.

whyboris:
Closes #204 

I decided not to hide the button (not an intuitive experience when you can't find what you expect to be there) -- there's just a message now in the _clips view_ gallery that says "no clips extracted" 👌 

ps - code-block moved down, but nothing changed except the `*ngIf`

### DIFF
--- a/src/app/components/home/home.component.html
+++ b/src/app/components/home/home.component.html
@@ -1010,31 +1010,6 @@
         ></app-file-item>
       </ng-container>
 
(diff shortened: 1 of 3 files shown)


cal2195:
Much better idea! 😄

PR #227, opened 2019-07-27
Element ref

whyboris:
`getElementsByClassName` is faster than `querySelector` 🚀 

Though it's shaving off milliseconds :trollface: 

https://gomakethings.com/javascript-selector-performance/ 🤷‍♂


cal2195:
Woop! 🎉

PR #228, opened 2019-07-27
Rescan buttons added to toolbar

### LINKED ISSUE #15: Extract Thumbnails on Startup
It would be nice to have the option to start extracting missing thumbnails when you open the app! 😄

whyboris:
Closes #15 

I decided against rescanning on startup option (see comment: https://github.com/whyboris/Video-Hub-App/issues/15#issuecomment-515683230)

Instead, simply allow users to add two of the rescan buttons to the toolbar ✅ 

This should handle 99% of the use cases without any added complexity & unexpected behavior.

While _some_ users might really prefer a rescan on startup, I think needing to click 1 more button after starting the app is a small-enough hassle to avoid the potential headaches for most other users.

ps - obviously two new icons will be added before `2.0.0` release 👍


cal2195:
Much simpler handling too! 👍

Issue #102, opened 2019-01-11
Last steps before v2.0.0 release


whyboris:
This issue is to keep a list of tasks that should be done after all the new features we'd like to see in version 2.0.0 are added, but before the release occurs. Most of these tasks are on top of the issues tagged for `Milestone v2.0.0`

This is a good place to see just how much needs to be done to make the current _master_ release-ready. In no particular order:

_Visual_
- [x] Reorganize the settings menu & tabs to be more coherent
- [x] Icon overhaul #51 
- [x] Confirm settings defaults - make the first app launch only display relevant buttons
- [x] Clean up the _Miscellaneous view settings_ - group and describe buttons better
- [x] CSS consolidation: pull colors and some re-used elements from a central file 
- [x] reevaluate colors - maybe better contrast for _a11y_ (accessibility) and harmonize all the filter and tag colors
- [x] CSS consolidation: some components use the same CSS, so, reuse those to keep DRY
- [x] confirm animations look good; add animations on newly added features

_Dark Mode_
- [x] Move the dark mode button to just-before-last button in the toolbar
- [x] Confirm app responds to OS X _Dark Mode_ settings / changes
- [x] Check the *dark mode* is working on all components; possibly simplify the CSS

_Code_
- [x] Streamline the view switching code - how element widths / heights are recomputed & how ng-scroll receives the expected heights / number of elements (we now have 6 views, used to be 3)
- [x] Remember each view's zoom level #16 
- [x] Consolidate the view-switching bug (some views currently don't show)
- [x] Check that all views handle window resizes and search-bar hiding
- [x] Make sure that adding a rating, scrolling away, and scrolling back persists the rating in the view
- [x] Remove most `console.log` statements
- [x] Add new keyboard shortcuts (for new views) & update the list in settings
- [x] Remove jankiness from the resolution, time, and rating filters

_Misc_
- [x] Update to latest Angular (v8)
- [x] Stress test the app with more than 1,000 videos in a hub

⚠️ Note -- unfinished tasks moved to #234

cal2195:
I think the full view scrolling issue has something to do with:
https://github.com/rintoj/ngx-virtual-scroller#items-with-variable-size

I tried enabling unequal child sizes, but the detail view started acting badly! 😢

whyboris:
The only view that has variable sizes is the _full-view_

[...]

PR #233, opened 2019-07-28
Cleaning still

whyboris:
- remove `randomizeGallery`
- remove `shouldWeShuffleNow` 
- make keyboard binding into `switch`/`case` rather than `if`/`else if`
- fix sort by year 🎉 

Some more work towards #102


src/app/components/common/settings-buttons.ts
```diff
@@ -68,7 +68,6 @@ export const SettingsButtonsGroups: string[][] = [
     'showMoreInfo',
     'compactView',
     'fontSizeLarger',
-    // 'randomizeGallery', // TODO - disabled for now
```
cal2195: 😢

whyboris: The `shuffle current results` button is still there -- the _randomizeGallery_ button, if I remember right, was a way to ask VHA to always show a random thumbnail from the extracted ones, rather than the first screenshot.

Given that we're storing the extracted screenshots separately from the first, it would require the app to load the larger .jpg strip rather than the single preview -- and it would require I write some more code that I don't expect many would use 🤷‍♂

PR #241, opened 2019-08-03
remove prediction, improve wizard

whyboris:
🤷‍♂ 

Prediction of time required to extract thumbnails and final _vha-folder_ size was outdated ... 

While it's nice and kind to manage expectations, there's a high likelihood that the estimate will be wrong. If it's an underestimate, users might think the app is slow. If it's an overestimate, they might think the app is slow. 

Now that users can re-scan (resume importing), there should be less importance in stating a correct import time at start 😅 

ps - this PR also improves the _wizard_ with a nice gradient that indicates you should scroll 🤔

### DIFF
--- a/src/app/components/home.component.html
+++ b/src/app/components/home.component.html
@@ -1238,6 +1238,8 @@
     <app-icon class="close-modal-icon" [icon]="'icon-close'"></app-icon>
   </div>
 
+  <div class="gradient-to-indicate-scrolling"></div>
+
   <div class="left-side" @slowFadeIn>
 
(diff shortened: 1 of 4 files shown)


cal2195:
If we want to be super fancy, we could calculate something along the lines of:

`remainingTime = (timeElasped / filesDone) * filesLeftToDo`

And update it as we go!

whyboris:
That is so clever 🎉 🎉 🎉 
I'll look into it 👍

whyboris:
@cal2195 thank you for the idea 🙇 
Implemented in #256 -- shows time remaining in both stages (meta- & screenshot- extraction)

PR #242, opened 2019-08-04
fix for saving regenerated previews when hub folder is not found

kaer:
I found a bug when regenerating thumbnails, the process finished but nothing was saved because the screenshot folder was missing.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -810,6 +810,10 @@ function verifyThumbnails() {
   // resume extracting any missing thumbnails
   const screenshotOutputFolder: string = path.join(globals.selectedOutputFolder, 'vha-' + globals.hubName);
 
+  // verify if the output folder existence
+  // if not found, recreate in order to regenerate image previews
+  verifyScreenshotOutputFolder();
+
   const indexesToScan: number[] = missingThumbsIndex(
     lastSavedFinalObject.images,
     screenshotOutputFolder,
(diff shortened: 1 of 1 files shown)


whyboris:
Thank you very much for your pull request! : slightly_smiling_face :

Could you describe how I could reproduce the bug first, before fixing it?

There is already a time when we perform `mkdirSync`, see: https://github.com/whyboris/Video-Hub-App/blob/master/main.ts#L552

I guess this can happen if the user ever creates a _Hub_ (and the app creates a _.vha_ file and a _vha-_ folder) and the user then _moves_ the `_.vha_` file somewhere else without moving the accompanying folder. Is this how you found the bug?

kaer:
I accidentaly removed the vha-<HUBNAME> folder, not the .vha2 file, and file previews didn't work for the hub, I tried to recreate in "verify all videos have thumbnails", the process started to recreate the previews but nothing was saved even processing every video.

In order to reproduce the bug use an existing hub and rename the vha-<HUBNAME> folder used for screenshots, open VHA and  try to recreate the previews in "verify all videos have thumbnails" option, check for new previews, they won't exist. Create a new folder with vha-<HUBNAME> and repeat the process, they'll exist.

whyboris:
I guess it may be worth including a `RECREATE FROM SCRATCH` option button -- if a user ever accidentally deletes / misplaces the `vha-` folder. 

Perhaps I should add a warning message when the `vha-` folder is missing -- when starting a `.vha2` file -- this way it's clear to users of the app what happened. Re-creating a full copy of a `vha-` folder isn't something I'd like to happen unless the user explicitly wants (and clicks on an appropriate button).

Thank you @kaer for identifying a possible problem and providing a solution. Before we implement a solution, let's agree on an approach. Your thoughts?

I'm open to comments / ideas.

kaer:
You are right, the user need to know what happened because this procedure in some cases may take minutes.

I have a feature that may be useful, group videos in collections defined by tags or manually. I have some other ideas too.

whyboris:
Ideas are welcome -- I recommend you create an [_Issue_](https://github.com/whyboris/Video-Hub-App/issues) for each individual idea you have 🎉 

Version 2 is coming soon, so new ideas would have to be included in future releases 👌 -- but let's discuss them 🤝

kaer:
Ok, thank you very much!

cal2195:
> I guess it may be worth including a RECREATE FROM SCRATCH option button -- if a user ever accidentally deletes / misplaces the vha- folder.

This is pretty much what the method you disabled is for! 😉 Although, I think adding in @kaer's solution wouldn't be a bad idea, maybe with a dialog to notify the user what's about to happen!

PR #247, opened 2019-08-05
Improve css, better breadcrumb navigation, & remove unused code

whyboris:
Hey @cal2195 -- in this PR I'm removing 3 lines of code you've added 😅 

I'm unsure what they do 😕 

Originally comes from this PR: https://github.com/whyboris/Video-Hub-App/pull/166

in the `onEnterKey` method:
```ts
      this.filters[origin].array = [];	
      this.filters[origin].bool = !this.filters[origin].bool;	
      this.filters[origin].string = '';
```

Please let me know if you have a moment and happen to remember 🙇

### DIFF
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -1373,24 +1373,15 @@ export class HomeComponent implements OnInit, AfterViewInit {
    */
   onEnterKey(value: string, origin: number): void {
     const trimmed = value.trim();
-    // removes '/' from folder path if there
-    // happens when user clicks folder path in file view
-    if (trimmed[0] === '/') {
-      this.filters[origin].array = [];
-    }
     if (trimmed) {
       // don't include duplicates
       if (!this.filters[origin].array.includes(trimmed)) {
         this.filters[origin].array.push(trimmed);
         this.filters[origin].bool = !this.filters[origin].bool;
         this.filters[origin].string = '';
+        this.scrollToTop();
       }
-    } else {
-      this.filters[origin].array = [];
-      this.filters[origin].bool = !this.filters[origin].bool;
-      this.filters[origin].string = '';
     }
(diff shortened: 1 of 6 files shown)


src/app/components/home.component.ts
```diff
@@ -1367,24 +1373,15 @@ export class HomeComponent implements OnInit, AfterViewInit {
...
     if (trimmed) {
       // don't include duplicates
       if (!this.filters[origin].array.includes(trimmed)) {
         this.filters[origin].array.push(trimmed);
         this.filters[origin].bool = !this.filters[origin].bool;
         this.filters[origin].string = '';
+        this.scrollToTop();
       }
-    } else {
-      this.filters[origin].array = [];
```
whyboris: I can't quite figure out what these three lines are meant for 😅 

@cal2195 -- please let me know if you remember 👍 

ps - currently, if nothing is entered in the search field and the user presses `enter` it clears out all the current search filters for that field (which I don't think is a desired behavior).

whyboris:
@cal2195 -- please let me know if you think I should return the 3 lines of code that you added (see comment) which I removed with this PR 🤝

src/app/components/home.component.ts
cal2195: That was the desired behaviour! 😂 I'll need to run the app again, but I think it was a quality of life improvement... 🤔

whyboris: I _do_ have the behavior of using the backspace button to clear out the last-entered search term, so pressing it a few times is enough to clear out everything 🤷‍♂ 

I think I accidentally pressed `Enter` once and it cleared out my search and I didn't like 😝

I wonder if `Escape` would be a good middle ground. Not as in-reach as the `enter` key, but more-expected as UX 🤷‍♂

PR #250, opened 2019-08-08
Open video at time code

### LINKED ISSUE #249: Feature request: Open video player with time code as parameter
Some video players like VLC or PotPlayer support proving the timecode as a parameter. It would be really helpful if it would be possible to do that from Video Hub App (e.g. when using the video strip).

whyboris:
Closes #249 

- Windows: only works with VLC, MPC, and POT players
- Mac: only works with VLC (note that `macVersion` boolean must be set to _true_ for this to work)

Electron has trouble opening the executable on a Mac, hence the hardcoded conditional in the `handleClick` method 😅 
https://github.com/electron/electron/issues/9490#issuecomment-520159654


cal2195:
This is very cool! 👍 I actually added this in one of my branches, but decided against it as the in app player allows for more control (using the sheet etc. to scrub through the video, not just at the start!)

But I like having both options! 😄 Could also add a custom exec command for users who know what they're doing? 🤔

PR #253, opened 2019-08-12
clip at half resolution option

whyboris:
🎉 Video can now be extracted at half resolution 🎉 

This should be very welcome for when extracting screenshots at high resolutions 🤤

### DIFF
--- a/main-extract.ts
+++ b/main-extract.ts
@@ -268,6 +268,7 @@ const generateScreenshotStrip = (
  * @param screenshotHeight   -- resolution in pixels (default is 100)
  * @param saveLocation -- folder where to save jpg files
  * @param clipSnippets -- number of clip snippets to extract
+ * @param clipHalfRez  -- if cilp should be half the resolution of thumbnails
  */
 const generatePreviewClip = (
   pathToVideo: string,
@@ -276,7 +277,8 @@ const generatePreviewClip = (
   screenshotHeight: number,
   saveLocation: string,
   clipSnippets: number,
-  snippetLength: number
+  snippetLength: number,
+  clipHalfRez: boolean
 ) => {
 
   return new Promise((resolve, reject) => {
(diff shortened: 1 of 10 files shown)


whyboris:
I tested at various resolutions -- but didn't try many file types -- I'm afraid _something_ might sometimes fail 🤷‍♂ 

@cal2195 if you have, off the top of your head, any concerns, please comment sometime 🙇

main-extract.ts
```diff
@@ -276,7 +277,8 @@ const generatePreviewClip = (
   screenshotHeight: number,
   saveLocation: string,
   clipSnippets: number,
-  snippetLength: number
+  snippetLength: number,
+  clipHalfRez: boolean
```
cal2195: fo shiz

main-extract.ts
```diff
@@ -300,7 +302,8 @@ const generatePreviewClip = (
       concat += '[' + (current - 1) + ':V]' + '[' + (current - 1) + ':a]';
       current++;
     }
-    concat += 'concat=n=' + (totalCount - 1) + ':v=1:a=1[v][a];[v]scale=-2:' + screenshotHeight + '[v2]';
+    concat += 'concat=n=' + (totalCount - 1) + ':v=1:a=1[v][a];[v]scale=-2:' +
+              (clipHalfRez ? screenshotHeight / 2 : screenshotHeight) + '[v2]';
```
cal2195: I assume ffmpeg will handle it, but have to tried halving an _odd_ number?

cal2195:
One thought - on a large library, people might have _small_ resolutions and _large_ resolutions. In this case, halving the small rezzzzzes would be annoying, but not halving the large ones would be bad too - maybe having a `MAX_HEIGHT` would be better instead?

main-extract.ts
whyboris: Currently I've hardcoded these heights:
```
'144', '216', '288', '360', '432'
```
No other ones are allowed. Halving any of them seems to work fine when I did spot-testing in the app 😅

whyboris:
I'm unsure I can picture what you're describing exactly. Please let me know if the suggestion below misses the point.

I guess there's no difficulty in having a separate selector for the _clipResolution_ rather than just a dumb _boolean_ toggle for _halfRez_ VS the default full resolution. 

So I'm thinking rather than `MAX_HEIGHT` we'll just have the `ScreenshotSettings` have `clipHeight` in addition to `height` 🤷‍♂

whyboris:
Improved functionality with #265 -- can manually select clip height 👍

PR #254, opened 2019-08-12
Duplicate finder

whyboris:
A really simple one, but better than nothing 😎


cal2195:
Very cool! I'll have to try this out! 😄 

Now if only I implemented the file management functions... 🤔

PR #265, opened 2019-08-24
select clip height independently

whyboris:
Better than a _boolean_ to simply select _half height_ 😉 

ps - a super-slick animation for the "current step" in the wizard (look at the numbered items getting a triangle!) 😎


whyboris:
@cal2195 -- no need for review -- just wanted to let you know I decided to implement this as per your comment about the `clipHalfRez` setting I put in in another PR: https://github.com/whyboris/Video-Hub-App/pull/253#issuecomment-521974198

PR #269, opened 2019-08-28
Handle corrupt & unfinished download videos

whyboris:
Summary: adding a timeout and killing any `ffmpeg` process that takes too long 😅 

New workflow:
1. do not do any kind of checking for whether the video file is corrupt
    - previous script takes too long (2-6 seconds for 15 screenshots -- even if the video is not corrupt)
2. spend at most 1s** trying to generate _first thumbnail_, if it fails, move on to the next file*
    - it takes about 60-80ms on my computer to extract one screenshot; 1s should be enough 🤷‍♂ 
3. spend at most 5s** trying to generate _filmstrip_, if that fails, move on to the next file*
    - this time will be a function of number of screenshots to extract, not yet determined
4. spend at most 10s** trying to generate _clip_, if that fails, move on to the next file*
    - this time will be a function of number of screenshots to extract, not yet determined

\* basically assume everything is corrupt and skip file (do not try to extract anything else)
\** better defaults will be experimentally found before merging 😅 

⚠️ I'm disabling the script that is meant to detect errors in files doesn't seem to do that -- perhaps it worked previously but currently it doesn't for me 😓 

I'll be working on this -- version 2.0.0 can't get released without handling corrupt files 🙆‍♂ 

ps - _promises_ are tricky to format, I'm following this: https://github.com/airbnb/javascript/issues/216#issuecomment-88066358

### DIFF
--- a/main-extract.ts
+++ b/main-extract.ts
@@ -72,7 +72,7 @@ const extractSingleFrame = (
   fileHash: string,
   screenshotHeight: number,
   duration: number
-) => {
+): Promise<boolean> => {
 
   return new Promise((resolve, reject) => {
 
@@ -80,49 +80,25 @@ const extractSingleFrame = (
       return resolve(true);
     }
 
-    const t0: number = performance.now();
-
     const ssWidth: number = screenshotHeight * (16 / 9);
 
-    const args = [
-      '-ss', duration / 10,
+    const args: string[] = [
+      '-ss', (duration / 10).toString(),
       '-i', pathToVideo,
-      '-frames', 1,
+      '-frames', '1',
       '-q:v', '2',
       '-vf', scaleAndPadString(ssWidth, screenshotHeight),
       saveLocation + '/thumbnails/' + fileHash + '.jpg',
     ];
-    // console.log('extracting clip frame 1');
-    const ffmpeg_process = spawn(ffmpegPath, args);
 
-    const killProcessTimeout = setTimeout(() => {
-      if (!ffmpeg_process.killed) {
-        ffmpeg_process.kill();
-        console.log('Single screenshot process KILLED!');
-        return resolve(false);
-      }
-    }, 300); // On my computer it routinely takes 60 - 80ms to extract
-
-
-    // Note from past Cal to future Cal:
-    // ALWAYS READ THE DATA, EVEN IF YOU DO NOTHING WITH IT
-    ffmpeg_process.stdout.on('data', data => {
-      if (globals.debug) {
-        console.log(data);
-      }
+     // On my computer it routinely takes 60 - 80ms to extract
+    spawn_ffmpeg_and_run(args, 300, 'single screenshot')
+    .then(success => {
+      return resolve(success);
+    })
+    .catch(err => {
+      console.log('what error?');
     });
-    ffmpeg_process.stderr.on('data', data => {
-      if (globals.debug) {
-        console.log('grep stderr: ' + data);
-      }
-    });
-    ffmpeg_process.on('exit', () => {
-      const t1: number = performance.now();
-      console.log('single frame: ' + (t1 - t0).toString());
-      clearTimeout(killProcessTimeout);
-      return resolve(true);
-    });
-
   });
 
 };
@@ -147,11 +123,9 @@ const generateScreenshotStrip = (
   screenshotHeight: number,
   numberOfScreenshots: number,
   saveLocation: string,
-) => {
+): Promise<boolean> => {
 
   return new Promise((resolve, reject) => {
-
-    const t0: number = performance.now();
 
     if (fs.existsSync(saveLocation + '/filmstrips/' + fileHash + '.jpg')) {
       console.log('thumbnails for ' + fileHash + ' already exist');
@@ -161,7 +135,7 @@ const generateScreenshotStrip = (
     let current = 0;
     const totalCount = numberOfScreenshots;
     const step: number = duration / (totalCount + 1);
-    const args = [];
+    const args: string[] = [];
     let allFramesFiltered = '';
     let outputFrames = '';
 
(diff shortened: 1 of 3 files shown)


whyboris:
@cal2195 -- I'm happy to get your input on this, but here's a small investigation I did:

```ts
const { performance } = require('perf_hooks');
const t0: number = performance.now();
const t1: number = performance.now();
console.log('SS check: ' + (t1 - t0).toString());
```
With 15 screenshots to extract per video, I see the check taking ~1 - 6 seconds per video (even non-corrupt videos), while extraction of the _filmstrip_, when the video is not corrupt routinely takes about 1~4 seconds. 

As a first-pass solution I think it makes sense to simply kill any _filmstrip_ extraction process that takes N seconds (where N is maybe 5 seconds? probably make N grow as a function of number of screenshots to extract). This would speed up the extraction, and prevent spending too long on the corrupt files 🤷‍♂ 

I really like the idea of the check, but it's taking long and is not useful for 99% of the clips people will deal with 😅 

Let me know what you think 👂

cal2195:
I like where you're going with this, I think it is a better idea than what I had implemented! 😄 6s per file was a small price to pay to actually finishing importing my library at all, but a timeout should work better!

The only issue I have it that not all computers are as fast, and not all file systems are quick too...

I, for example, run the importing stage of vha in a Linux vm on a server, with the files served over nfs - both would impact the speed of ffmpeg, which might leave me with files not imported properly.

Also, file types probably come into play too...

Testing a tweaking will be needed! 🙂

cal2195:
Looking at your notes, point 2 worries me the most over a network file system...

cal2195:
With a few debug outputs of failed files, I'm happy to run the changes over my full library and see what doesn't make it through! 👍

whyboris:
Thank you for the feedback 🙇 

I aim to give enough padding for extraction time (e.g. 300ms when as few as 100 are needed, though I see that networking can eat up quite a bit). I'm mildly considering (though it's a bit scary 😅 ) whether to run some kind of a monitoring script that can dynamically adjust the timeout a bit 🤷‍♂ 

I'm going to be testing this some more and I'm all ears for better sane defaults for timeouts 🙇

whyboris:
Experimentally, for 15 screenshots at highest resolution, 3-second clip at medium resolution:
- thumb: 160ms - 340ms
- filmstrip: 1000ms - 5400ms
- clip: 300ms -  1000ms
- clip thumb: 60ms - 100ms
I'll test more, leaving the above for posterity.

ps - I'm extracting from a 7200rpm HD onto a SSD on a pretty fast machine 😅 

As a rule of thumb, I'm thinking set max duration to _triple all the numbers_ and call it a day 😅

main-extract.ts
```diff
@@ -507,63 +412,150 @@ export function extractFromTheseFiles(
...
-      '-y', '-i', newFile,
-      '-vf', scaleAndPadString(width, height),
-      oldFile,
-    ];
+    // const t0: number = performance.now();
 
     const ffmpeg_process = spawn(ffmpegPath, args);
+
+    const killProcessTimeout = setTimeout(() => {
+      if (!ffmpeg_process.killed) {
```
whyboris: I'm unsure about this -- there's likely an official way to check if a process is still running 🤷‍♂

whyboris:
Merging -- I suspect it's decent-enough 🤷‍♂ 
I'm still unsure whether to implement something tricky like keeping an average of how long it took to successfully extract each type of thing ... feels like it could be very buggy.

Another approach would be to simply have another button to "add missing thumbs" in the settings that is more-thorough (by taking 2x as long) -- seems like the easiest solution 🤔 - unsure

cal2195:
Just an update on this - try this out recently I'm finding about 10-15% of new videos I add are now not completing thumbnail extraction. I'll investigate further, but it's a regression on my end. :(

whyboris:
Yeah -- I'm setting a timeout and killing any process that lives too long. This will be a problem on slower computers. 

I made the timeout about 3x of what it took to import video screenshots, but I'm guessing that's not long enough.

I was considering adding a second button that will scan in videos with a timeout that is twice as long as this one, but I'm unsure if that's the way to go ... 😓 

The big trade-off is that if I make the timeout longer, corrupt files / unfinished downloads take up that much longer to do. Perhaps a hybrid solution where It scans things quickly like this, and when done, it will re-scan with 2x the timeout 🤷‍♂  ... 

I had removed the functionality you put in that did a pre-check to see whether all screenshots are there, because the pre-check seemed to take as long if not longer per video as the actual screenshot extraction 😕 

I'm open to ideas / suggestions 🙇  -- please feel free to open an _Issue_ and link to here 🙇

whyboris:
@cal2195 -- curious -- for the 10-15% of the videos you describe, do they have the first thumbnail and are missing the filmstrip or something worse than that?

The current extraction process doesn't try to generate the filmstrip if the thumbnail extraction fails (could happen on corrupted or incompletely downloaded videos). 

If you click the "add any missing thumbnails", does it look like it tries to extract for a while, or fails out quickly? If it's fast, it's likely failing on the first screenshot; if it's slow, I suspect your computer isn't generating the filmstrip quickly enough and it simply times out (the feature I added in this PR) 😅 

If you have a moment, please clarify how the extraction failed 🙇

cal2195:
I've just tried for a complete recreation of the library there - your comment about the missing first thumbnails is correct, more of them are here now that it's had a chance to do the full extraction from the beginning. 👍 

There's still a few missing ones though, most notably they seem to be high quality videos (greater than 1080p), so they probably do take longer to process. Running "add missing thumbnails" just goes slowly, but doesn't appear to fix them so far... (are the timeout different currently?)

whyboris:
Thank you for the valuable feedback! 

My formula for making the timeout longer is linear, while resolution doubling is likely closer to a squaring of one of the factors. My VHA also timed out once on a 4K video at highest screenshot height at 30 screenshots. I'll adjust the formula before release to return timeout lengths mostly-strictly above the current results.

whyboris:
Fixing it with this commit: https://github.com/whyboris/Video-Hub-App/pull/288/commits/0cd073d88ac9be74eb32912ca901aab4ddd48e69

🤷‍♂ Not very happy about the timeout business, but perhaps it's better to extract more screenshots at the risk of taking much longer on corrupt files :/

PR #275, opened 2019-09-04
prevent sleep when extracting files

whyboris:
Hope it works 🤷‍♂ 

https://electronjs.org/docs/api/power-save-blocker

also adding a check to turn off the app gracefully when computer is about to shut down:

https://electronjs.org/docs/api/power-monitor


whyboris:
😅  annoying to test -- still haven't gotten a chance to see if it works as expected 🤞

whyboris:
👌  tested -- seem to work great 👍

PR #280, opened 2019-10-04
Fixed directory path and language abbreviation in i18n readme

R3l4x3:
- The path to the i18n directory was wrong
- The abbreviation in your example did not match the one in your link.


whyboris:
Thank you for the catch 🙇 
I resolved the merge conflicts as I merged another PR right before this one 👌

PR #285, opened 2019-10-12
should not interfere when installing next to version 1.X.X

whyboris:
Maybe it's only on my computer, but when installing video hub app 2.0.0 next to video hub app 1.3.0 it would corrupt version 1. And also after installing 2, installing 1.3.0 would install in the same folder `Video Hub App 2` 😖  ... 

Anyway, now I explicitly set the `appId` in _electron-builder.json_ and that should resolve the problem.

I tested on another computer 🤷‍♂ ... my computer still has some trouble 🤷‍♂

### DIFF
--- a/electron-builder.json
+++ b/electron-builder.json
@@ -1,5 +1,7 @@
 {
   "productName": "Video Hub App 2",
+  "appId": "com.videohubapp.videohubapp2",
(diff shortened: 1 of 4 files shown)


whyboris:
Will test again when I'm making the final build and testing on different computers 👌

PR #301, opened 2019-11-04
Allow folder with `_` prefix to be scanned

### LINKED ISSUE #293: valid folders ignored with leading underscore (all platforms)
https://github.com/whyboris/Video-Hub-App/blob/c20940ba586da0007e973af610253641c865af94/main-support.ts#L242

It would be nice if the ignore patterns were configurable.  I had multiple files missing and realized anything in a directory which had a name with a leading `_` was being skipped.  Use of a leading underscore helped with the natural sort order in some instances.   If the concern is metadata/previews on mac systems in `__MACOSX` folders, using that full folder name as an ignore pattern might be an option to avoid files unintentionally being skipped on all platforms :-)

cal2195:
Closes #293. 👍 

In a future PR, we could make this configurable! 😄

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -239,8 +239,8 @@ export function getVideoPathsAndNames(sourceFolderPath: string): ImageElement[] 
 
   const finalArray: ImageElement[] = [];
   let elementIndex = 0;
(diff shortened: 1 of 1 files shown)


whyboris:
🙇 Thank you 🎉

PR #327, opened 2020-01-02
allow to set favorite thumbnail

### LINKED ISSUE #310: Define favorite thumbnail
🤷‍♂ a user emailed me with this request ... maybe useful to others?

Currently, you can drag & drop any image you'd like over any thumbnail and the thumbnail will be replaced.

It looks like the user is interested in selecting one of the many generated thumbnails a particular thumbnail. 

I'm thinking the easiest way would be:
1) hover over the thumbnail until the particular thumb you like shows up
2) right click
3) select "make default" (though I'm unsure what short phrase/word would be good to put there -- I don't want it to be long

Another alternative would be in the "Details" view (right click -> details / ctrl + click / top-right icon when hovering over a thumbnail in thumbnail view) where you see all the screenshots individually -- perhaps hovering over a thumbnail you can click on it to add a star on it -- to indicate it's the "primary preview thumbnail"

Discussion welcome 👍

whyboris:
⭐️ Go to the details modal and click on the star to set favorite thumbnail!

Closes #310 🚀

### DIFF
--- a/src/app/components/views/thumbnail/preview.component.ts
+++ b/src/app/components/views/thumbnail/preview.component.ts
@@ -105,7 +105,7 @@ export class PreviewComponent implements OnInit, OnDestroy {
     }
   }
 
-  mouseIsMoving($event) {
+  mouseIsMoving($event: any) {
     if (this.hoverScrub) {
       const cursorX = $event.layerX;
       this.indexToShow = Math.floor(cursorX * (this.video.screens / this.containerWidth));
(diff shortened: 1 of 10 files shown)


src/app/components/views/thumbnail/preview.component.ts
```diff
@@ -88,7 +97,9 @@ export class PreviewComponent implements OnInit, OnDestroy {
       clearInterval(this.scrollInterval);
     }
 
-    if (this.returnToFirstScreenshot) {
+    if (this.video.defaultScreen) {
```
cal2195: Should this not also take into account `this.returnToFirstScreenshot`?

whyboris: Good catch -- I'll look into it and make another commit tomorrow (I'm away today) 🤝

whyboris: Thank you Cal 🙇‍♂️ -- I think it's fixed 👍

Issue #318, opened 2019-11-27
Tag tray needs scrollability


whyboris:
An email come in -- seems like at least one person has filled up their custom tags to the brim 😁 

Let's make a container (create a new one if needed) that holds the tags scrollable 👍

navin1551:
Hi, this is my first time contributing to an open source project and would like to take a shot at this. Can I try to add the scrollability?

cal2195:
Sure thing @navin1551! Go ahead! 😄

whyboris:
Thank you @ScottBWar for fixing this issue 🙇‍♂️ #328

PR #328, opened 2020-01-05
Add a scrollbar to the bottom tag tray

ScottBWar:
CSS Changes to add a working scrollbar to the bottom tag tray

Changes to address:
https://github.com/whyboris/Video-Hub-App/issues/318

### DIFF
--- a/src/app/components/home.component.html
+++ b/src/app/components/home.component.html
@@ -1350,19 +1350,19 @@ title="{{ 'TAGS.closeTagTray' | translate }}"
     {{ 'TAGS.presentTags' | translate }}:
   </span>
   <button
-    (click)="manualTagShowFrequency = !manualTagShowFrequency"
-    class="show-frequency"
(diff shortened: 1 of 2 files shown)


whyboris:
Thank you for the improvement!

There's a bit of a regression:

2020-01-05 14_52_18-Window

While we want to fit more tags into the view, I don't want it to feel cramped, so I'd prefer to retain the spacing I had (above the tags; fitting 4 rows completely and aligning the "tags in this hub" with the first row of the tags).

I'm happy to merge this in and fix it up a bit -- just let me know if you'd prefer to make the changes yourself 🙇

ScottBWar:
I think I overlooked the spacing a bit while moving things around to make sure the scrollbar showed up. I'll tweak it now so it's closer to what you had before.

ScottBWar:
@whyboris spacing should look better now

PR #334, opened 2020-01-06
[#332] Fix bug of tags that persists after switching hubs

### LINKED ISSUE #332: Tags in tag tray persist when switching to another hub
An unwanted bug 😅 

1) open a hub with some tags
2) open the tag tray (see the tags)
3) switch to another hub (that has no tags) via the "recently opened" feature
4) see the same tags

Unsure if occurs in other scenarios. I suspect it's a simple fix -- just have to trace out where I forgot to reset the tags service 😅

jamessspanggg:
Fixes #332 

```
When switching over one hub to another, the tags in the tag 
tray persists, as the tags of the previous hub are not cleared 
after switching over to the next hub.

This PR clears the tags whenever users switch between hubs,
which will prevent the previous tags to persist.
```

### DIFF
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -660,6 +660,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
 
       this.folderViewNavigationPath = '';
 
+      this.manualTagsService.removeAllTags();
       this.setTags(finalObject.addTags, finalObject.removeTags);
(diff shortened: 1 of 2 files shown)


jamessspanggg:
I've noticed another issue where adding new tags while the tag tray is open, the tag tray does not update to reflect the latest added tags.

PR #337, opened 2020-01-15
Fixes for incorrect/missing resolution when audio stream is first in the ffprobe output

kevinseelbach:
Fixes for #335 as discussed


kevinseelbach:
Oops.. pushed those commits with wrong SSH key

PR #338, opened 2020-01-15
Fixes for incorrect/missing resolution when audio stream is first in the ffprobe output

kevinseelbach:
Fix issue #335 as discussed

I just went with the max screen width value instead of calculating _width * height_ , multiplying them together seemed like overkill for now.. 

Thanks @cal2195 your suggestions worked out great.

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -386,6 +386,17 @@ export function extractAllMetadata(
 }
 
 /**
+ * Iterates ffprobe output to find stream with the best resolution (just width, for now)
+ *
+ * @param metadata  the ffProbe metadata object
+ */
+function getBestStream(metadata) {
+  return metadata.streams.reduce(function (a, b) {
+    return a.width > b.width ? a : b;
+  });
+}
+
+/**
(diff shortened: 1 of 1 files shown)


main-support.ts
```diff
@@ -385,6 +385,17 @@ export function extractAllMetadata(
...
+/**
+ * Iterates ffprobe output to find stream with the best resolution (just width, for now)
+ *
+ * @param metadata  the ffProbe metadata object
+ */
+function getBestStream(metadata) {
+  return metadata.streams.reduce(function (a, b) {
+    return a.width > b.width ? a : b;
+  });
+}
```
kevinseelbach: Eh, I realize now this looks a little funny pulled out into its own function, when it's a one-liner for the most part, but it's less dense this way and if the calculation gets more complicated this is better.

whyboris:
Finally tested with some video files that previously showed up as `0x0` resolution and they all show up with correct resolution 🎉 🙇  🤝 

Thank you @kevinseelbach  😁

Issue #348, opened 2020-01-28
Show full thumbnail for single parent folders


danjool:
![solo-thumb-folder](https://user-images.githubusercontent.com/10146364/73228385-a6745200-4133-11ea-8dfc-100c26495de2.png)

It seems more natural to me that if folder only has one media item, then why not show the whole thumbnail?  

I have this implemented locally already, but either github is being ornery or there is some kind of push access required.  Also I could see where this would be a View Settings toggle button, as different people would have different preferences.

cal2195:
Hi @danjool,

To make a pull request, fork this repo, push your changes to your forked repo and then create a pull request from there! 🙂

danjool:
Thanks for the heads up!
https://github.com/whyboris/Video-Hub-App/pull/350

whyboris:
Thank you @danjool for the improvement -- I think it looks much better!

Sorry took 2 weeks for me to get to merge in your change 🙇 

Closes #350

PR #350, opened 2020-01-28
solo folder thumb

danjool:
a folder with only 1 media element shows the full size thumbnail

### DIFF
--- a/src/app/components/views/thumbnail/preview.component.html
+++ b/src/app/components/views/thumbnail/preview.component.html
@@ -17,13 +17,25 @@
       (click)="videoClick.emit($event)"
       [ngStyle]="{ height: imgHeight + 'px' }"
     >
+
+      <!-- Show one thumbnail if there is only one -->
       <div
-        *ngIf="folderThumbPaths[0]"
-        class="folder-item"
+        *ngIf="!folderThumbPaths[1] && folderThumbPaths[0]"
+        class="folder-item-fullsize"
         [ngStyle]="{
           'background-image': 'url(' + folderThumbPaths[0] + ')'
         }"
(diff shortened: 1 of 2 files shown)


whyboris:
Thank you for the Pull Request! I'll see how it looks later this week -- seems like it could look good 👍

src/app/components/views/thumbnail/preview.component.html
```diff
@@ -17,13 +17,25 @@
       (click)="videoClick.emit($event)"
       [ngStyle]="{ height: imgHeight + 'px' }"
     >
+
+      <!-- Show one thumbnail if there is only one -->
       <div
-        *ngIf="folderThumbPaths[0]"
-        class="folder-item"
+        *ngIf="!folderThumbPaths[1] && folderThumbPaths[0]"
```
whyboris: FYI, since you're treating this as an `if / else`, an approach you can do with _Angular_ is similar to line 12 above
https://github.com/whyboris/Video-Hub-App/pull/350/files#diff-2dc1e1902eca723eb5d284832a0502d4R14
`<el *ngIf="condition; else selector">` and then `<el #selector>` 

I'll likely update the code into this pattern after merging 👌 

Thank you for the excellent contribution 🤝

whyboris:
Closes #348

src/app/components/views/thumbnail/preview.component.html
whyboris: Just a minor change: https://github.com/whyboris/Video-Hub-App/pull/356/commits/2be2f1ba659ded26b4bea4ec0667418a4a903e9a 

Cheers 🙂

PR #351, opened 2020-01-28
flatten-folder

danjool:
adds rt click context menu with folder flattening action

### DIFF
--- a/main-support.ts
+++ b/main-support.ts
@@ -285,6 +285,60 @@ export function getVideoPathsAndNames(sourceFolderPath: string): ImageElement[] 
 }
 
 /**
+ * Walks the folder structure, starting potentially from a SUBfolder of the hub (folderViewNavigationPath),
+ * moving all media contents that VHA can recognize to the starting subfolder folderViewNavigationPath.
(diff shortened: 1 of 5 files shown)


whyboris:
Really cool! I'll have to try it out first, but seems like it could be very useful to some 👍

whyboris:
Sorry this has stalled 😓

I'm still thinking about this 🙇  -- I've not forgotten 😅

danjool:
All good.  I still think the UX of the feature could be better, just not sure what that looks like.

whyboris:
I'm still thinking about this feature -- I really like the idea (I have to use [suction](https://www.ghacks.net/2009/06/09/consolidate-files-with-suction/) to get the same functionality otherwise). 

I'm considering adding a "tree view" (no ETA, maybe 2022 😓 ) of all the folders as its own dedicated side-bar (so users can move videos around using the app interface). If that feature arrives, this feature would be quite intuitive (and there would be visual feedback about what happened to the "tree" folder structure) 🤝 

Cheers 👍

whyboris:
I ended up creating a `CLI` / script that performs this action: https://github.com/whyboris/flatten-this-folder

I'm eager to get `3.3.0` out - I'll think about including the _flatten folder_ functionality in a future release 👍

PR #361, opened 2020-02-13
update ko.json

mmega:
+fuzzy search,  +right-click 'delete',  +playlist, 
and Word replacement

line 262
I think it's possible


i18n/ko.json
```diff
@@ -268,110 +280,112 @@
...
-    "noSuchFileFound": "해당파일을 찾을 수 없습니다. :",
-    "pleaseChangeName": "허브이름을 변경하십시오.",
-    "selectDefaultPlayer": "선호하는 비디오 플레이어를 선택하십시오.",
-    "selectPreviousHub": "이전에 저장된 허브파일을 선택하십시오.",
-    "selectRootFolder": "소스비디오 폴더 선택",
-    "videoFolderNotFound": "소스비디오 폴더를 찾을 수 없습니다. :"
+    "continueAnyway": "무시하고 계속",
+    "directory": "디렉터리",
+    "doesNotExist": "존재하지 않습니다.",
+    "HubAlreadyExists": "같은 이름의 Hub가 있음",
```
whyboris: this got accidentally capitalized `h` -> `H` -- I think it needs to be exact, else it will not work

i18n/ko.json
```diff
@@ -1,265 +1,277 @@
...
+    "fileUnion": "동영상 조합",
+    "folder": "폴더 이름",
+    "folderUnion": "폴더 조합",
     "found": "찾음",
-    "magic": "폴더 또는 비디오 이름",
-    "or": "또는",
-    "recentlyOpened": "최근에 생성한 허브 :",
-    "similarTo": "다음과 유사합니다.",
-    "sortAlphabetical": "알파벳순",
+    "fuzzy": "퍼지 검색",
```
whyboris: I have no problem with adding this line, but I don't think it will be used anywhere in the app at this time

i18n/ko.json
```diff
@@ -1,265 +1,277 @@
...
+    "toggleFileName": "파일 이름, 해상도, 길이 표시",
+    "toggleMinimalMode": "최소 모드 전환(모든 패널 숨기기)",
+    "toggleSearchBar": "검색 사이드바 표시",
+    "toggleSettings": "설정 표시",
+    "toggleWizard": "마법사 표시"
   },
   "SIDEBAR": {
-    "and": "그리고",
-    "dupeByHash": "중복의심",
+    "and": "and",
```
whyboris: You would know better -- whether `그리고` or `and` is better 👌

i18n/ko.json
```diff
@@ -268,110 +280,112 @@
...
-    "openPrevious": "이전허브 열기",
-    "openPreviouslySaved": "이전에 저장된 비디오 허브 열기",
-    "recentlyOpened": "최근에 열었던 허브 :",
+    "createVideoHub": "동영상 Hub 만들기",
+    "demoVersion": "데모 버전은 50개 동영상으로 제한",
+    "eachSnippet": "각각의 조각들을",
+    "expected": "Expected:",
+    "extractClips": "클립 미리 보기 생성",
+    "found": "found",
+    "HubDirectory": "(Hub 디렉터리)",
```
whyboris: Accidentally `h` -> `H` ... needs to be lowercase first letter

i18n/ko.json
```diff
@@ -268,110 +280,112 @@
...
-    "oneScreenshotEvery": "스크린샷 1개마다",
-    "openPrevious": "이전허브 열기",
-    "openPreviouslySaved": "이전에 저장된 비디오 허브 열기",
-    "recentlyOpened": "최근에 열었던 허브 :",
+    "createVideoHub": "동영상 Hub 만들기",
+    "demoVersion": "데모 버전은 50개 동영상으로 제한",
+    "eachSnippet": "각각의 조각들을",
+    "expected": "Expected:",
+    "extractClips": "클립 미리 보기 생성",
+    "found": "found",
```
whyboris: Should these be in English? (this and `expected`, `minimum`, `any`, etc above) ?

whyboris:
Thank you very much for the translation!

I spot-checked some strings and they look good! I'm trusting the rest are similarly good -- thank you for your work 👍 

There are some strings that were in Korean but now are in English, is that on purpose?

i18n/ko.json
mmega: It's my mistake. hub -> Hub

I changed all the hubs on the left. Hub ->hub

Use the Hub on the right?
I think it would be nice to use the name of the Video-Hub-App.

i18n/ko.json
mmega: No problem. It's okay to keep using it

i18n/ko.json
mmega: and, or
I thought English would be good.

i18n/ko.json
mmega: It's my mistake. hub -> Hub

I changed all the hubs on the left. Hub ->hub

i18n/ko.json
mmega: On my level, it's hard to explain. So even if you don't have enough explanation 
,please understand.
I could see this sentence in the auto tag tray.
[containing any 4 minimum] the translation of this sentence 
In Korean, it is natural to be in the order of [minimum 4 any containing].

[339 line "filterList": "Containing"] requires to enter ['minimum' instead of {Containing}].
[342 line "minimum": "Minimum"] requires to enter ['Containing' instead of {'minimum'}].
{any, 4} -> {4, any} must be.  Unnatural `5stars`
So I couldn't change it.

Sometimes it's not natural if a sentence consists of multiple lines.
The wizard's case is similar. (2,3,4,5) Unnatural `2stars`

It is natural that the sentence of [select video, hub directory] 
should be in the order of [video, hub directory select].
So I used [select (video,hub directory).

Wizard again - import new hub - 4.2
{369 line "oneScreenshotEvery": "1 screenshot every",} -> {"oneEveryScreenshot": "","}, turn it into a blank
{367 line "minutes": "minutes",} -> {"minutes": "minutes 1 screenshot every",} can I change it to translate it?
Can I change the other parts as above?

I wanted to change it naturally, but it was because of my low level.

I should have done it better, but I couldn't.
I'm sorry I've uploaded in a hurry 
I'll check back and update

whyboris: I'll trust whatever you decide on -- I'm aware languages structure things differently, and words like 'or' and 'and' may be widely understood too (you would know better).

Thank you for your work 🙇 

No rush 👍

whyboris:
Moved over to #371 👍

Issue #365, opened 2020-02-16
Backlog 2.X.X


whyboris:
Miscellaneous stuff that could be done someday:

### Not Done
- [ ] Run `lint:scss` again 😓 (also see #196)
- [ ] Migrate to `eslint` (since `tslint` will be [deprecated soon](https://github.com/palantir/tslint#tslint)) 
    - probably just copy this: https://github.com/maximegris/angular-electron/pull/410
- [x] Add setting: _Folder view_ show _random_ screenshots rather than first 4.
   - I don't want to set random without an option to disable: some users _might_ find it unpleasant to see folder thumbnails change every time they navigate
- [ ] Check scrolling isn't janky or is making items disappear too soon (as _full view_ used to before, possibly still). Investigate if there are too many repaints happening with dev tools. Reconsider the element shadow - might be slowing things down [ref talk about 60fps](https://addyosmani.com/blog/making-a-site-jank-free/)

Continuation of the previous backlog: #284 

Other ideas welcome 👍

whyboris:
I'm thinking of creating a release for `2.2.0` within the next 5-ish days 🎉 

I think 2 more issues would be good to address before then (I'll fix these; comments welcome):
- Update the colors on the manually-added tags in the details view
    - they are currently blue, even though in the tag tray they show up as yellow 😕 
- Word cloud shows manually-added tags, but when you click on them, the string is added to the _file_ search and not the _tag_ search (and so no video may be found) 😅

whyboris:
I'm updating the _colors_ with #384 but don't feel like fixing the word-cloud bug; will leave that for later 🤷‍♂ (no one has ever emailed me about it, so it's probably a bug rarely encountered, and I'd rather release the next version of the app 👍 ).

LordJon:
Aah is that why the word-cloud tags gives no results sometimes. I assumed it was somerthing in my settings but haven't looked into it yet so I now avoid using the feature hahaha.

whyboris:
Screenshot for `README.md`

![video-hub-app](https://user-images.githubusercontent.com/17264277/82097107-3ed91700-96d0-11ea-8679-87fa3e07cd0b.jpg)

orkomlosh:
_Add setting: Folder view show random screenshots rather than first 4.
I don't want to set random without an option to disable: some users might find it unpleasant to see folder thumbnails change every time they navigate_

so I implemented this ⬆️  . see PR #502 

[...]

PR #384, opened 2020-03-03
remember scroll offset when returning to root in folder view

### LINKED ISSUE #382: Scroll back to original location when clicking 'UP' inside folder
1. When in _Folder View_ scroll down some distance, 
2. open a folder,
3. click the `up` button
4. notice that the app resets the scroll to the top of the page

It would be nice if it went back to roughly where you were before clicking on the folder.

This feature was requested by a user. Seems useful when navigating using folders.

Feels like it may be easy to implement: 
- for simplicity the feature could be available only for the _root_ folder
- simply store the y-offset (or whatever number associated with it) in the component
- make _virtual scroller_ scroll to there
    - it may be possible to use `scrollToIndex` or `scrollToPosition` in _virtual scroller_ to achieve this

whyboris:
Closes #382 

Updates colors for tags (see below)
 - no more green tags 😅

### DIFF
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -288,6 +288,9 @@ export class HomeComponent implements OnInit, AfterViewInit {
 
   deletePipeHack: boolean = false; // to force deletePipe to update
 
+  // crappy hack to remember where to navigate to when in folder view and returning back to root
+  folderNavigationScrollOffset: number = 0;
+
   // ========================================================================
   // Please add new variables below if they don't fit into any other section
   // ------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


src/app/components/home.component.ts
```diff
@@ -1098,6 +1106,31 @@ export class HomeComponent implements OnInit, AfterViewInit {
...
+   * Scroll appropriately after navigating back to root folder
+   *
+   * Rather hacky thing, but works in the basic case
+   * Fails if user enters folder, changes some search or sort filter, and navigates back
+   */
+  scrollAppropriately(filter: string) {
+    if (filter === '') {
+      // TODO -- change `1` to `0` once the bug is fixed:
+      // https://github.com/rintoj/ngx-virtual-scroller/issues/420
+      this.virtualScroller.scrollToPosition(this.folderNavigationScrollOffset, 1);
```
whyboris: Will need to change this to `0` once the bug is fixed:
https://github.com/rintoj/ngx-virtual-scroller/issues/420

whyboris:
Fixing colors (see #365) 

image

Suggestions welcome 👂

PR #402, opened 2020-04-14
Quick file rename

byrw:
Fixes #400

PR #407, opened 2020-04-19
Styling for the renaming file feature

whyboris:
Thank you to @byrw for the code: #402

This PR is just minor CSS fixes to make the feature more-harmonious in its look with the rest of the app 🤝

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -321,6 +321,7 @@
     "continueAnyway": "Continue Anyway",
     "directory": "Directory",
     "doesNotExist": "does not exist",
+    "fileRenameError": "File was not renamed. Some error occurred.",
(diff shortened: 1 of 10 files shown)


whyboris:
As a first step (just to get my orientation) I have reset the styles back to what they were before the migration of code to a new component 🤞

whyboris:
The main plan is that the title will always be editable in the _Details view_ 🚀 
When the title is changed, a message will show up below saying "press 'Esc' to cancel renaming", otherwise (pressing "Enter" or leaving the text input field) the file will be renamed.

Whether successful or not, a confirmation message will briefly appear to let the user know 🧷

whyboris:
I think it's ready 😅

PR #409, opened 2020-04-23
Drag n drop tags

whyboris:
👍 😎 👍 

You can drag tags from the tag tray onto any video and the tag will be added 😁


whyboris:
Ready 🚀 

Was easier than expected 😁

Issue #394, opened 2020-04-04
Stop initial scan for total number of videos after 500 videos found


whyboris:
Sometimes users want to create a hub from a directory with thousands of files. During the initial "wizard" the app freezes as it tries to find the exact number of videos. 

I think it would be a better experience if the app reported the exact number of videos up to a point (for example 500 videos), after which it reported "over 500 videos found" 👌 

This would make the first step never crash 😅

deepank120896:
@whyboris Is anyone working on this?

whyboris:
No -- please feel free to fix 👍 

I'm thinking even if we detect 1 video it's enough to say "videos detected" -- please feel free to implement as you would like.

Please let me know if you'd like any advice / help / direction 🤝

whyboris:
@deepank120896 -- let me know if you're intending to work on this 🤝

whyboris:
I'll add this feature before release: #412

whyboris:
I'm thinking this feature will not be needed once #383 lands 😁

whyboris:
I'm leaving this issue open just in case someone, who like me, is also annoyed at the app freezing 😅 

This feature will be fixed with #443 - `Video Hub App 3` release 🎉 

No need to work on this issue 👌

whyboris:
Closing for cleanup. Will be released with #443 👍 -- already fixed in the `video-hub-app-3` branch 😎

PR #412, opened 2020-04-30
New version release

whyboris:
I'm hoping to release more often.
Let's get this released 🎉 
Maybe make it public this weekend 😉 

fixes:
- manual tags have enough space above for the ❌ to delete the tag
- details view 2 updates when tag is added / removed
- makes `showDetails2` view work on first open
- 'check for new release' button in settings 🎉


whyboris:
I'll be adding another change before `2.2.3` gets released 👍 
I want to fix #394 🚀

PR #417, opened 2020-05-10
Refactoring

whyboris:
Cleaning up the very overgrown `home.component.html` 😅 by moving template stuff to newly-created & dedicated components:
- shortcut component
- ribbon component
- sort-order component
- tag-tray component
- wizard component 😅 

Reducing the length of `home.component.html` template from *1,872* lines of code to *1,340* 😓 

A later `css` cleanup is in order 😓 - will do after PR merges


whyboris:
@cal2195 -- no need to review -- adding you just FYI -- creating a bunch of component files and moving code away from `home.component.html` since it's overgrown 🌱 🌱 🌱 

I should have done this from before version `2.0.0` but kept being lazy 😜

PR #419, opened 2020-05-12
New components

whyboris:
Finishing off the refactoring - moving code out of `home.component.html` template to make it more-manageable 😅 

New components:
- rename-modal component
- resolution-filter component
- star-filter component
- recently-opened component
- similar-tray component

`home.component.html` was 1,073 lines-of-code-long and is now 951 🎖️ 

Not a drastic reduction in lines but having dedicated components makes the codebase a bit more organized and easier to refactor/clean up later 😅


whyboris:
@cal2195 -- adding you just FYI -- no need to review 👌

Issue #386, opened 2020-03-10
Let users see clip thumbnail same as regular thumbnail


whyboris:
Currently the preview thumbnail for the clips is different than that for the main gallery preview. This was done on purpose, but I could see how someone would prefer the same screenshot in both. 

This would probably be a super-easy thing to toggle in the settings 👍

whyboris:
Thank you @martincaron for implementing this feature (#422) 🎉 🙇  🥇

PR #422, opened 2020-05-13
Let users see clip thumbnail same as regular thumbnail (#386)

martincaron:
Issue https://github.com/whyboris/Video-Hub-App/issues/386 - Let users see clip thumbnail same as regular thumbnail

- Added new toggle under `View Settings` ->` Clips View`
- Setting only appears in settings menu, not in ribbon
- If enabled,  it uses the same thumbnail as the other views
- Works with Folder view or Single Clip
- If enabled, it also respects the `returnToFirstScreenshot` setting


whyboris:
Thank you for this PR 🤝 
Code seems great -- I'll test the changes today and will merge before releasing 2.2.3 🎉

whyboris:
And thank you for the _angular_-specific `(mouseout)` fix 😉 (replacing the `onmouseout` event)

Issue #399, opened 2020-04-09
New filter aspect ratio


Aer20:
It would be helpful to have a new filter for the aspect ratio of the video
Name in filter : range of aspect ratio
Vertical 16:9 : 0 - 0,64
Vertical 4:3 : 0,65-0,89
Square : 0,90-1,10
4:3 : 1,11 - 1,59
16:9 : 1,60 - 1,89
Cinema 2:1 : 1,90 - 2,10
Cinema 2,35:1 : 2,11 - 20,00

whyboris:
This is an interesting idea. Visually it would be too cluttered to fit it as a horizontal slider in the sidebar, but perhaps it could become a "sort by" feature where it sorts videos from vertical to horizontal or from horizontal to vertical 🤷 

As a filter it may be relatively easy to add to the current setup -- for every video compute the ratio (width/height) and sort by that number.

Will accept an "Aspect ratio" sort PR 👍

whyboris:
Thank you @martincaron for implementing this feature (#423)  👍 🎖️ 🤝

PR #423, opened 2020-05-14
Sorting option for aspect ratio (#399)

martincaron:
Added sorting option for aspect ratio. ( Issue #399 )

- New setting option under `Search Settings` -> `Sorting options in the dropdown`
- For every video compute `aspectRatio` = `ImageElement.width` / `ImageElement.height`
- Implemented Ascending and Descending order
- Division by Zero doesn't seem possible because we have `AllowedScreenshotHeight` type upstream preventing a height of `0`, but I can add some checks to be safe. Feedback ?


whyboris:
Thank you for the PR 🙇 
Code looks good -- I'll test locally and likely merge in before releasing 2.2.3 🎉 

Minor notes (no need to change, just food for thought):
- you can use ternary `return decreasing ? 1 : -1` as a shorter way of writing what you have
- we can make the sorting faster (I'll check on 10,000 video hub if it's slow) by pre-computing the ratios in the `insertTemporaryFields` see: https://github.com/whyboris/Video-Hub-App/blob/master/main-support.ts#L93 
    - if we do this we should then remember to `insertTemporaryFields` https://github.com/whyboris/Video-Hub-App/blob/7c7b40e0ccf049abe6c581c605cee197f8f330b3/main-support.ts#L233
    - with 10,000 videos when trying to sort by a value there are going to be super-many comparisons, and not having to divide numbers would be much faster 🤔 

👋 🙂 👍

martincaron:
All makes sense ! 👍
I should be able to make these changes

whyboris:
@martincaron -- I tested things out on 10,000 videos in a hub and the sorting is so fast I don't think we need to optimize anything.

If you think it may be worth doing, please feel free to open a new PR. I'm eagerly merging this one as it concludes the version `2.2.3` release :tada

PR #449, opened 2020-06-10
add file size slider filter

litelime:
So I basically just copied the duration slider to use fileSize property instead. 
The icon would need to be changed and maybe some ui fix so the size value isnt overlapping the slider bar.


whyboris:
Thank you @litelime for this excellent addition 🏅 

It will be released in version 3.0.0 (the next release) whenever I'm done with the multi-folder hub feature (several months from now I think 🤞 )

PR #451, opened 2020-06-14
add rmvb and asf file support

frez777:
More info:
https://fileinfo.com/extension/asf
https://fileinfo.com/extension/rmvb


whyboris:
Looks great! And thank you for the alphabetical order 👍

Issue #438, opened 2020-05-25
Allow delete bypassing recycling bin


whyboris:
#397 - users are having trouble deleting over network drives

This is because I'm using [trash](https://github.com/sindresorhus/trash) to send files to the _Recycling Bin_ / _Trash_ instead of the destructive `rm path/to/file.mp4` command 😅 

Since making it work over network drives requires some setup, let's instead do this:

### What to do:
1. Add a new button below the "Show 'Delete' option in the right-click menu"
2. Make the button a boolean that makes the app send over the boolean value to `delete-video-file` method in `main.ts` which will
3. Run `rm path/to/file.mp4` 👌

jamesd35:
Hi can I work on this?

whyboris:
Thank you @jamesd35 for the feature addition 🙇 

Solved with #453 -- will be released along with v3 later this year 🤝

PR #453, opened 2020-06-14
Allow delete bypassing recycling bin #438

jamesd35:
Added toggle button in settings for permanent deletion bypassing trash

### DIFF
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -300,6 +300,8 @@ export class HomeComponent implements OnInit, AfterViewInit {
 
   tagTypeAhead: string = '';
 
+  dangerousDelete: boolean = false;
+
   // ========================================================================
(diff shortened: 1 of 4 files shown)


whyboris:
Code looks great! I'll test it out in the app sometime (this weekend probably) and merge it in 🚀

src/app/components/home.component.ts
```diff
@@ -300,6 +300,8 @@ export class HomeComponent implements OnInit, AfterViewInit {
 
   tagTypeAhead: string = '';
 
+  dangerousDelete: boolean = false;
```
whyboris, line 303: Maybe this was WIP stuff -- this variable isn't needed as you can simply use the _settingsButtons_ `dangerouslyDelete.toggled` boolean state when sending the message through `ipcRenderer` (around line 1910).

I'll remove after merging 👌

whyboris:
Thank you @jamesd35 for fixing this issue #438 🏅

PR #443, opened 2020-05-27
Video Hub App 3 🎉

whyboris:
Huge thank you to @cal2195 for his *Async Continuous File Importing* 🎉 #383

This PR is for the branch `video-hub-app-3` which is an integration branch of many different changes.

Very many changes came in from #432 - read the (incomplete) list there 😅 
More changes came in via #444 😁 
Persistent watching and cleaner code via #460 😎 
Add missing thumbnails & rescan once via #465 🚀 
Show when folder not connected; delete old screenshots via #466
Now you can delete old screenshots for videos not in gallery & remove videos that are no longer on the HD via #467 
Add comments to any video via #483 😎 

Still lots to do:
- [ ] implement `blocks` (see Cal's [comment](https://github.com/whyboris/Video-Hub-App/pull/432#issuecomment-634632286))
- [ ] fulfill Cal's [*TODO LIST*](https://github.com/whyboris/Video-Hub-App/pull/383#issuecomment-632051168) 🤞 
- [x] clean the new extraction code
- [x] remove or archive old extraction code
- [x] allow selecting more than one folder as 'source folder'
- [x] delete `postinstall-web.js` and `postinstall.js`

🤞 let's hope this goes well 😅 🚀

PR #458, opened 2020-06-22
Support video player launch arguments

CrendKing:
Partially support launching video player with arguments. Currently only load the arguments from `videoPlayerArgs` key in settings.json. UI needs to be implemented.

### DIFF
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -863,7 +863,7 @@ export class HomeComponent implements OnInit, AfterViewInit {
 
       const execPath: string = this.appState.preferredVideoPlayer;
 
-      this.electronService.ipcRenderer.send('open-media-file-at-timestamp', execPath, fullPath, this.getVideoPlayerArgs(execPath, time));
+      this.electronService.ipcRenderer.send('open-media-file-at-timestamp', execPath, fullPath, this.getVideoPlayerArgs(execPath, this.appState.videoPlayerArgs, time));
     } else {
       this.electronService.ipcRenderer.send('open-media-file', fullPath);
     }
@@ -874,24 +874,21 @@ export class HomeComponent implements OnInit, AfterViewInit {
    * @param playerPath  full path to user's preferred video player
    * @param time        time in seconds
    */
-  public getVideoPlayerArgs(playerPath: string, time: number): string[] {
+  public getVideoPlayerArgs(playerPath: string, playerArgs: string, time: number): string[] {
     // if user doesn't want to open at timestamp, don't!
-    if (!this.settingsButtons['openAtTimestamp'].toggled) {
-      return [];
-    }
+    const argz: string[] = playerArgs ? playerArgs.match(/[^\s"']+|"([^"]*)"/g) : [];
 
-    // else, figure out the correct command line flags
-    const argz: string[] = [];
+    if (this.settingsButtons['openAtTimestamp'].toggled) {
+      if (playerPath.toLowerCase().includes('vlc')) {
+        argz.push('--start-time=' + time.toString()); // in seconds
 
-    if (playerPath.toLowerCase().includes('vlc')) {
-      argz.push('--start-time=' + time.toString()); // in seconds
+      } else if (playerPath.toLowerCase().includes('mpc')) {
+        argz.push('/start');
+        argz.push((1000 * time).toString());          // in milliseconds
 
-    } else if (playerPath.toLowerCase().includes('mpc')) {
-      argz.push('/start');
-      argz.push((1000 * time).toString());          // in milliseconds
-
-    } else if (playerPath.toLowerCase().includes('pot')) {
-      argz.push('/seek=' + time.toString());        // in seconds
+      } else if (playerPath.toLowerCase().includes('pot')) {
+        argz.push('/seek=' + time.toString());        // in seconds
+      }
     }
 
     return argz;
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -214,6 +214,7 @@
     "changeLanguage": "Change language",
     "checkNewVersion": "check for new version",
     "chooseVideoPlayer": "change",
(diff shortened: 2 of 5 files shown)


whyboris:
Thank you for the PR 🤝 

I have 0 clue what command line arguments could be used for 😅 . 

If you have a moment, could you share why you pass in arguments into your video player? Besides opening at timestamp I don't know of any other use case at this time.

CrendKing:
For example, I have a hub that contains movies. I usually play movies in fullscreen, and I use MPC-HC/MPC-BE. To open the player in fullscreen mode, I need to pass `/fullscreen`. Also, if I want the player to automatically close once the playback is finished (instead of just closing the file), I need to to pass `/close`. These arguments are commonly used in other apps that launch video players, such as [streamlink-twitch-gui](https://github.com/streamlink/streamlink-twitch-gui/blob/master/src/config/players.json) (it is a frontend for streamlink, specifically designed for [Twitch](https://www.twitch.tv)).

whyboris:
Really interesting!

If you want per-hub launch arguments, it seems like it shouldn't be too challenging: store it in the .vha file (add a property in the _interface_ and make sure it's saved and used in the right place) 🤔 

If it's per-hub setting, in the UI it might make sense to store those in the "Statistics" settings tab (in version 3 this tab will be renamed to something else as it will be the place to manage the source folders - if your hub has more than one). 

One challenge I have with implementing a feature like the one your PR adds is that I don't know how it's used so I'll likely not implement it well / implement something else that's not useful.

From the looks of it, and from your description, it's just a matter of adding a text-box somewhere and saving the entered text somewhere. 

It seems like your PR did the back-end portion. I think I'm not opposed to adding the feature to the app. If you are interested in adding the front-end portion in, I'll be happy to merge the PR after. Don't worry too much about making it pretty -- just wedge in the elements wherever you think appropriate and I can mess with CSS to make it look more-harmonious with the rest of the app.

Side thought: I'm unsure whether a per hub or global setting is best. If you think both are useful, perhaps there's a way to do it -- a box that will apply to all hubs in the 3rd settings tab, but can be overridden per hub if a hub has its own setting via "Statistics" tab 🤔 

If you work on the feature, could you, for convenience, base it off the `video-hub-app-3` branch? And make the PR be against it too? https://github.com/whyboris/Video-Hub-App/pull/443

CrendKing:
I added a textbox for the arguments, under video player setting. It's only available if user chose "preferred video player". You might want to adjust the style yourself, or maybe add some label for it if the placeholder is not obvious enough.

src/app/components/home.component.ts
```diff
@@ -874,24 +874,21 @@ export class HomeComponent implements OnInit, AfterViewInit {
...
-  public getVideoPlayerArgs(playerPath: string, time: number): string[] {
+  public getVideoPlayerArgs(playerPath: string, playerArgs: string, time: number): string[] {
     // if user doesn't want to open at timestamp, don't!
-    if (!this.settingsButtons['openAtTimestamp'].toggled) {
-      return [];
-    }
-
-    // else, figure out the correct command line flags
-    const argz: string[] = [];
+    const argz: string[] = playerArgs ? playerArgs.match(/[^\s"']+|"([^"]*)"/g) : [];
```
whyboris, line 879: Could you explain why there is a `.match()` happening?

I would have thought whatever the argument(s) the user adds in the input would just be added verbatim 🤔 

I'm not very familiar with how arguments get used with video players, but I thought it would be less puzzling to users to have the exact string just added as is 🤔 

Please let me know 🙂  👍  🤝

whyboris:
Hey @CrendKing -- sorry took so long for me to get to this. I think what you have is great -- I just wanted to know about the _regex_ that's there -- I would have expected the exact string users add to be appended as a player launch argument 🤔 

Looking forward to merging it in 🤝

src/app/components/home.component.ts
CrendKing: The arguments is an array of string used to call [child_process.spawn()](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options). Substrings between spaces must be splitted into array elements to function. For example, if I want to spawn `player.exe /start /close "/jump 123"`, I need to call `spawn('player.exe', ['/start', '/close', '/jump 123'])` instead of `spawn('player.exe', ['/start /close "/jump=123"'])`, right? The regex is used to argument with spaces and quotes.

Sure, according to https://stackoverflow.com/questions/23487363/how-can-i-parse-a-string-into-appropriate-arguments-for-child-process-spawn, we can avoid these by using the `shell` option for `spawn()`, but I wanted to maintain the structure of the source code as much as possible. You guys probably had your reason to not use `shell`. If not, let me know or feel free to change yourself when merging.

whyboris:
Hey @CrendKing -- sorry -- I think because I merged my branch and renamed `master` to `main` this PR got closed 😱 

I didn't mean to close it -- I think it's great and I'm happy to merge it in!

Thank you for explaining the _regex_ -- as I said it's mostly the case that I don't know how command line arguments for video players work 😅 

I don't know the benefits of using `shell` over `spawn()` -- perhaps I tried one and it worked -- it's been a while since I added the feature.

If you think there's a benefit to using `spawn()` instead, please feel free to change as you recommend 🤝 

Cheers!

src/app/components/home.component.ts
whyboris: Thank you for the explanation! Please feel free to change `spawn()` to `shell` if that makes the code cleaner and/or has other benefits.

Please re-submit your PR -- it got closed by GitHub when I merged my branch and renamed `master` to `main` 😓

whyboris: ps - why not `.split(' ');` ? seems simpler if we're just splitting all the white spaces 🤔  or would this miss out on something?

whyboris:
Reopened as #487 🎉

PR #460, opened 2020-06-28
VHA 3 WIP PROGRESS

whyboris:
🤷 

Slowly making my way to release 🐌


node/main-extract-async.ts
```diff
@@ -107,47 +108,27 @@ export function sendNewVideoMetadata(imageElement: ImageElement) {
...
-      const element = cachedFinalArray[i];
-      if (element.partialPath === file.partialPath && element.fileName === file.name) { // consider comparing full path in case source folders overlap
-        found = true;
-        break;
-      }
-    }
-    console.log('found:', found);
-
-    if (found) {
+    if (cachedMetaExtracted.has(file.fullPath)) {
```
whyboris, line 115: @cal2195 -- thought you might like this minor rework 😉 

Rather than search through the array to see if `fullPath` matches I created a _map_ so it's just a matter of lookup `map.has(string)` 😎 

All this is still WIP -- but I wanted to share 👋

cal2195: How did we not think of this?! 😅 

Although I must say, VHA is one of the most performant applications I've ever used for the amount of data it processes on the fly! I wonder how much optimisation you get with pipes for free... 💭

whyboris: Yeah -- Angular pipes only re-render if any of their inputs change - so it's not the whole stack of pipes that runs when you change something - and many pipes just return whatever comes into them. And since it's not literally "loading" the data into the pipe - but just receiving a reference to a location, it can pass it along to the next pipe as if it was a single integer - fast 🚀 😎

tsconfig-serve.json
```diff
@@ -1,25 +1,23 @@
 {
   "compilerOptions": {
-    "sourceMap": true,
     "declaration": false,
-    "moduleResolution": "node",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
+    "lib": [
+      "es2020",
```
whyboris, line 7: @cal2195 -- if you happen to know what's going on here, I'd appreciate a comment -- I have almost no idea what I'm doing 😅 

I was thinking of updating things to latest libraries -- and everything seems to work, but the one that messed me up is the `target` (see line 12) -- "es5" -- when I bumped it up to `ESNext` or `ES2020` everything failed (`main.ts` errored out on the first line with an error, something like "can't use Import outside of module" 🤷 

I would have thought these things would be backward compatible ... though I may be thinking wrong about everything 🤷 I updated to Node 14 and it still didn't help -- so I'm unsure what's happening.

If you happen to know the about `target` thing and can comment, please. Otherwise, no worries -- everything is working fine 👌

node/main-extract-async.ts
whyboris: Also -- after looking at the code some more I couldn't figure out why we were ever looking whether a hash was present in previously-added files, so I removed this piece of logic.

Please let me know if I'm missing something / going to run into a case where it's needed later 😅

whyboris: I named one variable `alreadyInAngular` (for transparency about its function) -- and it gets populated with full paths to all the files in the currently-opened vha file. 

When a watcher finds a new file, it compares the full path to the list -- if it's on the list, we are done. If it's _not_ on the list, we `createNewElement` (and create a _hash_ for it of course).

When we send over the newly-created element to Angular, we update the `alreadyInAngular` with full path -- so if a subsequent watcher finds the same file it doesn't create a copy.

🤔  -- seems like it should work out fine.

🙇 thank you again for the amazing setup -- all I'm doing now is just cleaning up a few things and making it work with more than one input folder -- the functionality is all yours 🤝 🎉 🙌

Issue #401, opened 2020-04-12
Store portable app settings next to .exe


whyboris:
Windows portable stores the `settings.json` in 
`C:\Users\<user>\AppData\Roaming\video-hub-app-2`

It should instead store it right next to the .exe file (seems like a better behavior, right?)

Comments welcome. PRs welcome 🙇

whyboris:
Thank you @jamesd35 for fixing this 🙇 🤝 🏅

PR #461, opened 2020-07-01
Store portable app settings next to .exe #401

jamesd35:
Windows portable saves settings.json inside same directory as the executable.


whyboris:
Super cool! Will try it out this weekend 😁 

This will close #401 🎉

whyboris:
Thank you @jamesd35 for the fix! Sorry took me a while to get to it 🙇

PR #466, opened 2020-07-03
VHA3 - check if folder is live (connected)

whyboris:
👍 

- show indicator when folder is not connected on thumbnails & in "current hub" settings
- add "cleanup thumbnail folder" button - deletes all thumbnails not present in the hub any more 👍


node/main-extract-async.ts
```diff
@@ -310,3 +318,54 @@ export function extractAnyMissingThumbs(
     }
   });
 }
+
+/**
+ * Scan the output directory and delete any file not in `hashesPresent`
+ * @param hashesPresent
+ * @param outputDir
+ */
+export function removeThumbnailsNotInHub(hashesPresent: Map<string, number>, outputDir: string): void {
```
whyboris, line 327: @cal2195 -- thank you for implementing `chokidar` and `async` 🙇 

This combo makes me feel like a 10x coder 🤩 -- unsure if this is how you would have implemented the feature -- but I add a watcher to look through all the screenshots, and any that don't match an existing hash get sent to a queue runner (`deleteThumbQueue`) to get deleted 😁 

Nice and smooth -- and I think without bugs 😅

PR #467, opened 2020-07-05
VHA3 - remove old

whyboris:
😅 going to try to make it work: removing of videos from gallery that are no longer on HD 😓 

🤷 various cleanup as well as usual

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -17,7 +17,7 @@ const windowStateKeeper = require('electron-window-state');
 // Methods
 import { createTouchBar } from './node/main-touch-bar';
 import { setUpIpcMessages } from './node/main-ipc';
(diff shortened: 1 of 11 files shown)


whyboris:
There's likely still bugs because of my number/string kerfuffle:
I wanted to use a `Map<number, string>` - number pointing to the path of input folder
When I wanted to save it to `.vha2` file it would need to be converted to an object 👌 
Objects can have integer keys, but they are automatically treated as if they were strings
So the `Map<number, string>` when asked if it `.has('1')` would fail 😅 
I'll see if I can fix it up in later 🙆‍♂️ or perhaps it's already fixed with one weird trick `parseInt(key, 10)` :trollface:

PR #478, opened 2020-07-09
Webserver control api

TheTacoScott:
Proof of concept for review/discussion, I have no reason to expect this would be merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 /node_modules
-
+/.idea
 /dist
 /release
 /tmp
--- a/command-server.ts
+++ b/command-server.ts
@@ -1,0 +1,18 @@
+const express = require('express');
+const commandServer = express();
+const commandServerPort = 3000;
+const { ipcMain } = require('electron');
+
+
+
+
+module.exports = function(webContents) {
+  commandServer.get('/', (req, res) => {
+    webContents.send('command-server', {'action': 'CLEAR_WEBFRAME_CACHE'});
+    res.send(`Hello World`);
+  });
+
+  commandServer.listen(commandServerPort, () => console.log(`Command server listening on port ${commandServerPort}`));
+
+};
+
--- a/main.ts
+++ b/main.ts
@@ -113,6 +113,8 @@ function createWindow() {
   });
   mainWindowState.manage(win);
 
+  require('./command-server')(win.webContents);
+
   myWindow = win;
 
   // Open the DevTools.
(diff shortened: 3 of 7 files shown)


TheTacoScott:
Feels like this kind of thing could be enabled/disabled and generally configured via a command line arg.

--enable-command-server --command-server-port=9404 --command-server-address=0.0.0.0 
etc

TheTacoScott:
since it seems communication between the main proc and the renderer(s) is all done with ipc style json message passing via channels it seems like this could be overloaded in an abstract way that "passes through" the raw RPC call you send it via the web interface and it just sends it on the respective channel

Like:

0.0.0.0:3333/rpc/(main|renderer)/**channel_name**

payload: {args sent to channel}

Would make it very bare bones, but it would have access to a huge array of functionality

cal2195:
Just some ideas here:

I'd probably prefer to also have a toggle in-app, to allow for persistent toggling on or off the server. While using dev, I'd use the cli, but once I build the app, it would be annoying to have to launch it with the flag. 🎏 

From a security point of view, should we look into HTTPS, and an api key that need to be passed to control the app?
It could just be a random string generated and saved by the app, which it just checks against all incoming commands! 😄 
At the very least, maybe we should only bind to localhost by default, with an option to change for remote access! 👍 

The idea that this just interacts with the already existing IPC renderer is SUPER cool, but we'd better make sure there's some checks! `delete-video-file` _might_ have undesired consequences if done by an unauthorized party! 😉 But I think we should have a way to enable this as well, as it could be useful for some usecases! 👍

TheTacoScott:
https = with a self signed cert feels like it would cause more trouble than it's worth as if someone wants to expose this app to the internet they should really just handle that with a http proxy that they set up and maintain.

Although an API generated that must be passed in the header: I like this idea.

TheTacoScott:
At the very least, maybe we should only bind to localhost by default, with an option to change for remote access! == total agreement

If we have an API based "simple authentication" route, then you can fully offload the "checks" you mention to the person using the web server. It's an advanced mode, designed for odd use cases. A "use at your own risk" feature.

whyboris (approved):
I'm _loving_ it! 

I've been focused on a narrow set of capabilities for VHA and hadn't really thought about the _server_ aspect of the app. I tested this code and it works beautifully (other PC on my WiFi was able to get "hello world" 🎉 ) 

After release of `3.0.0` I'll experiment with some things and we'll almost surely have the option to enable a server in this app 😁

whyboris:
Very related: #603 got merged into `main` -- we now have a dedicated `server` file already set up with an `express` server and a decent UI 👍

whyboris:
@TheTacoScott -- please let me know if it's OK for me to close this PR as I think what I have is a good-enough start, similar to what you have but already in the `main` branch 🤝

whyboris:
Closing as #603 has added a `server` and the associated UI  🤝  thank you for your work 🙇

PR #481, opened 2020-07-11
Single-click / Double-click option

sezalmittal987:
added the feature of double-click/single-click while similar videos tray is open


whyboris:
Thank you for your contribution! 🤝 
Sorry took me 2 weeks to get back to you.

It behaves _beautifully_ 💮

I played around with the functionality and I might in the future just add a "single click" vs "double click" button in the Settings so that the behavior is always consistent regardless of whether the tray is open 🤔 I think some users might prefer double-clicking always 🤔 Please take no offense if I change / update some of your code 😅

sezalmittal987:
oh yes,it is a great idea!

whyboris:
Thank you @sezalmittal987 again for your contribution -- it makes implementing my #500 *double click mode* so much easier!

I hope you don't mind I renamed your variable `videoClickSimilar` to `doubleClick`: https://github.com/whyboris/Video-Hub-App/pull/500/commits/1f52c904146023b38c16b46f3cdf34fde7bbbd68

And I think I found a neat way of handling the detection between double-clicking and single-clicking:
https://github.com/whyboris/Video-Hub-App/pull/500/commits/ef5f9ec903a4c71897a27b9927be1b0302347bce

🤔 Rather than creating extra ternary conditionals in the template, I have a check inside the method that handles opening videos: if it's a double-click, it doesn't open the video, just updates the `currentClickedItemName` 😉

sezalmittal987:
oh it's actually cool.

whyboris:
I found another improvement/simplification: 

no need to have two different emits from the _thumbnail.component_ -- we can just re-use the `videoClick` event, but just include another boolean property `doubleClick` inside it 😁 

https://github.com/whyboris/Video-Hub-App/pull/500/commits/6928853a2a1bc88098e6d78aad378c4fef4f06b8

Issue #472, opened 2020-07-07
Add comments to each video


whyboris:
Very easy to do technically, just a matter of adding the UI for it.

I'll think about adding this feature after version 3 comes out 🤔

whyboris:
🚀 going to be in VHA 3 🎉 
Getting done here: #483

whyboris:
Fixed with #483 -- will be part of VHA 3.0.0 😎

PR #483, opened 2020-07-14
VHA3 - random fixes

whyboris:
🤷 WIP 👷 

New feature:
- add _video notes_ to any video
- allow searching through notes 👍 
- CSS to make notes look good 😓 kind-of-ok 🤷 

Will close: #472


whyboris:
Thanks for the catch Cal!

whyboris:
I've left this PR unattended for too long -- I'd rather merge it and continue with cleanup in a separate PR 😕 

The `finalArrayNeedsSaving` may be handled better in another way -- I'll see if maybe I don't need all the `editFinalArrayStars` and `editFinalArrayTag` and `editFinalArrayYear` emissions (`.emit($event)`) -- and I could just remove all of them and replace those and the "details changed" events with a single emit of `finalArrayNeedsUpdating` 👍

PR #487, opened 2020-07-27
Support video player launch arguments

CrendKing:
Add an input box in the settings page for video player launch arguments. The custom argument string along with automatically generated timestamp arguments are passed to the launched video player. The custom arguments are also saved in the "videoPlayerArgs" key in settings.json.


whyboris:
Thank you for recreating the PR I unintentionally closed #458 🤝

PR #491, opened 2020-08-03
Debugging - and fixing bugs 🎉

whyboris:
Turns out adding `ipcRenderer.on('message', ...)` in components that get created often just creates numerous listeners 😅 

Refactoring to only add the event listeners in the `home.component` and have every other component listen to a _BehaviorSubject_ 👌 

Solves my bugs 😎


whyboris:
Lost electricity to the house for a few days because of a storm 🤷 worry not - I'll be back soon 😉

Issue #355, opened 2020-02-09
thumbnails, tags


mmega:
thumbnails view - compact view
Sometimes can't see all thumbnails
Scroll down all see
![20200209_173024](https://user-images.githubusercontent.com/57275671/74099219-906b8600-4b64-11ea-962c-bacc20d3f917.jpg)

details view - show manual tags 
Times played 0 all visible, Times played 1 not all visible
![20200209_173929](https://user-images.githubusercontent.com/57275671/74099221-906b8600-4b64-11ea-8c21-9cf8628e4fdb.jpg)

Thanks

whyboris:
Thank you for opening this issue. I hope I can understand what the issue is; it seems like you're reporting 2 bugs?

1) In the first screenshot you do not see all thumbnails showing
    - (even though the view has more of them -- I can see the scrollbar size indicates there are more) 
    - I suspect once you start scrolling they all show up, is that correct?
    - The issue is _not_ about the screenshots all being black (as if no thumbnails were extracted), right?

This is a known (to me) issue (and it's very minor, I think)
    - It's only a bit annoying (since on-scroll everything appears) but is also probably easy to fix; 
    - would be great if someone could track down where in the code we need to fix it 
    - (I suspect it's a matter of forcing the virtual-scroller to update its view 🤷‍♂ )

2) In _details_ view, the app indicates that you have watched a video more than 1 time, but does _not_ indicate when you've never opened a video (through the app).
    - **This is intended behavior** (I programmed it this way on purpose)
    - Do you think it would be better if it always showed "_Played 0 times_" when the video hasn't been opened through the app yet?

Please let me know if I've misunderstood the issues.

mmega:
First of all, I'm using this software very well. 
I'm sorry i didn't have enough explanation. All thumbnails are normal. I covered it with a black box. 
I couldn't show all the thumbnails and letters.
I'll reattach it and explain it. 

1. Sometimes [thumbnails view-compact view] doesn't see all the thumbnails, so I wanted to let you know.
Attached a gif. 4.51mb
![20200210_133415](https://user-images.githubusercontent.com/57275671/74123004-8606c600-4c10-11ea-94d8-94038296ac84.gif)



2. I use auto-generated tags and manual tags. view it using manual tags.
see the video more than once, the tagbox, letters will be pushed down and truncated?cut off?
I just hope to see all the tags and letters.
[...]

PR #498, opened 2020-08-10
hide donut when ribbon is collapsed

orkomlosh:
@whyboris I got your comment very seriously: https://github.com/whyboris/Video-Hub-App/issues/355#issuecomment-671490094
the donut in this PR  will collapse with the top bar. works like charm.
if you want to do something else with the donut let me know.

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -431,6 +431,7 @@
     "sizeToImport": "mb of disk space",
     "stage": "stage",
     "stopImport": "stop",
+    "importInProgress": "Importing in progress",
     "timeToImport": "minutes to import",
     "videoDirectory": "video directory",
     "videos": "videos"
--- a/src/app/components/home.component.html
+++ b/src/app/components/home.component.html
@@ -21,7 +21,7 @@
   ></div>
 
   <!-- Progress indicator - "Donut" -->
-  <div *ngIf="importStage === 'importingScreenshots'" class="cancel-progress" @donutAppear>
+  <div *ngIf="importStage === 'importingScreenshots' && !settingsButtons['hideTop'].toggled" class="cancel-progress" @donutAppear>
     <span
       (click)="cancelCurrentImport()"
       [ngClass]="{ 'cancel-progress-dark': settingsButtons['darkMode'].toggled }"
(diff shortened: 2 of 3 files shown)


whyboris:
Thanks for the solution!

Hiding the doughnut progress bar isn't hard as you noticed ... my (unstated) worries were about how to communicate to the user that something is happening -- that we're still extracting things.

I've added a new indicator (green pulsating icon) next to the total number of files counter on the top of the sidebar (to indicate that we're scanning a folder for files) ... I think I'll add the same indicator to show we're extracting thumbnails (on top-right where the doughnut was) to indicate we're extracting screenshots 👍 

Let me know if you'd like to add the green icon yourself (see the `div` with `scanning-indicator` class) 🟢  -- else I'll merge in your PR and add it myself 👌 🤝 😁

orkomlosh:
ok, I'll take a look on this tomorrow,
BTW i noticed the the application header is also printing the the percentage and that it's indicating about the scan.

whyboris:
You're right! I forgot about that. It may be enough 🤔 

No rush on anything you do with VHA! All help is appreciated 👐

image
Seems like it could be nice to have.

ps - the taskbar should also already be indicating screenshot extraction progress 😎

orkomlosh:
So i tried to implement it and i manage to do something,
but i must admit i'm not very good with CSS, aligning and positioning objects.
(my background is mostly from .net)
so it's not pixel perfect: 😄 
image
 and i'm not sure what the right way to do it.
please correct me.

whyboris:
Super awesome! Thank you for the clean work 🙇 
I'll nudge the icon a few pixels with some CSS 👌

whyboris:
I nudged the icon a bit with this commit: https://github.com/whyboris/Video-Hub-App/pull/499/commits/1118d143475b9c9b877295a4a72fbddf56799daa
Also removed the container that was no longer needed (as I used `position: absolute`: https://github.com/whyboris/Video-Hub-App/pull/499/commits/9e4394b1bfb9e80d12add33c82d2dd9eb8a02187

It's hacks everywhere 😅 -- the right-padding on the `.all-settings` class is what makes the buttons start disappearing on the right as they get too close (and are about to overlap with the settings button and the newly-added indicator) 😆

orkomlosh:
Great, thx for the detailed explanation

PR #502, opened 2020-08-14
Random folder screenshots

orkomlosh:


src/app/pipes/folder-view.pipe.ts
```diff
@@ -62,19 +64,35 @@ export class FolderViewPipe implements PipeTransform {
...
+    return hashes;
+  }
 
-    if (hashes.charAt(0) === ':') {
-      hashes = hashes.slice(1);
+  extractRandomPreviewHashes(files: ImageElement[]): string {
+    let hashes: string = '';
+    for (let index = 0; index < 4; index++) {
+      const randomIndex = Math.floor(Math.random() * files.length);
+      hashes += ':' + files.splice(randomIndex, 1)[0].hash;
```
whyboris, line 94: 🤔 `.splice` removes elements _in place_ [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) 🤔 

> The splice() method changes the contents of an array by removing or replacing existing elements and/or adding new elements in place.

I think this _may_ be fine here, but I'd like your thoughts and a comment on this 🤔

whyboris: I suspect it's fine because I think we're working with a temporary copy and not at all messing with the original `finalArray` 😅 -- and I tested the code -- it looks good and behaves well, so I think we're clear ... but still would appreciate a comment 👍 

ps - This file was a tense algorithm to write 😅

src/app/common/settings-buttons.ts
```diff
@@ -475,6 +477,13 @@ export const SettingsButtons: SettingsButtonsType = {
     title: 'BUTTONS.playlistButton',
     toggled: false
   },
+  'randomizeFoldersScreenshots': {
+    description: 'BUTTONS.randFolderScreenDesc',
+    hidden: true,
+    iconName: 'icon-image',
```
whyboris, line 483: I wonder if this button would ever be useful in the _buttons ribbon_ 🤔 
I suspect the users would likely toggle it once and never fidget with it after.

I appreciate that you found an icon for the button, but what are your thoughts on removing this property (and the `svg`) - and the button will appear as a plain toggle in settings-only?

src/app/pipes/folder-view.pipe.ts
orkomlosh: Yeah,
So i got this thought myself when i wrote it. But than i got to the same conclusion that this is only a copy and this array is used in order to get the screenshots anyway.
It took me a while to think about how to do it. And this seems the most proficient way to get it done and get unique 4 hashes.

src/app/common/settings-buttons.ts
orkomlosh: Ok, yeah i think you are right.
I'll get to it later today.

orkomlosh:
I committed the change to remove the icon, PR is ready

Issue #445, opened 2020-05-28
main-filenames as configurable option


frez777:
I would like to do a feature request: **to have an option to add new file extension on-the-go** (I'm talking about the file main-filenames.ts which defines what kind of file the video hub will try to import).

Some people may have some video files that works with ffmpeg but its not imported by default because the file extension is not on main-filenames.ts file, some of them maybe not worth add for everybody like a `.part` file that may have unfinished downloaded file video that ffmpeg can work with, or a different/customized file extension not very used.

Another nice think would have an option to try to import every file doesn't matter the filename, of course with a warning on this option to people avoid using it without knowing what its about, could be very useful for video-only directories.

whyboris:
Seems doable 👌 

PR welcome. I imagine in the _main settings tab_ (3rd tab - the one that opens by default) below all the "select default video player" options 👍 

I'm guessing it would accept a comma-separated list of extensions (without periods), so
```
part, abc, xyz
```

Other suggestions welcome for how to handle it 👍

frez777:
I tried so far to implement the last thing a mention on this issue, which is an option to skip the extension check, because its easier and I'm not a developer, my skills are pretty bad. but unfortunately it didn't work.

What I tried was this:

---
```src/app/common/settings-buttons.ts```:
add:
```    'searchForAllFiles'``` on SettingsButtonsGroups constant

and
```
  'searchForAllFiles': {
    description: 'BUTTONS.searchForAllFilesDescription',
    hidden: false,
    moreInfo: 'BUTTONS.searchForAllFilesMoreInfo',
    title: 'BUTTONS.searchForAllFilesHint',
    toggled: false
  },
``` 
on SettingsButtons constant

---
```i18n/en.json```:

```
    "searchForAllFilesDescription": "Skip the file extension check before trying to import a file",
    "searchForAllFilesHint": "Skip the file extension check",
    "searchForAllFilesMoreInfo": "Beware that this option will make the import process try to import all files, even non-video ones, if unsure keep it desabled",
```

---
```main-support.ts```:

add an exportable variable:
```
export var skipExtensionCheck = false;
```

and changes the if responsible for the check:
```
if (skipExtensionCheck === true || (acceptableFiles.includes(extension.toLowerCase()) && !file.name.match(fileIgnoreRegex))) {
```

---
```src/app/components/home.component.ts```:

import the variables:
```
import { ImportStage, skipExtensionCheck } from '../../../main-support';
```

[...]

PR #507, opened 2020-08-17
Additional extensions setting

orkomlosh:
#445 
ok, so it wasn't that straight forward as i thought in the beginning it took me a while to understand how to implement it .
let me know what you think.


orkomlosh:
Accidentally closed the PR so i reopened it 🤦‍♂️

src/app/components/settings/settings.component.ts
```diff
@@ -33,7 +36,41 @@ export class SettingsComponent {
...
+    this.additionalInput = this.appState.addtionalExtensions;
+  }
+
+  applyAdditionalExtensions() {
+    if (this.isAdditionalInputValid()) {
+      this.appState.addtionalExtensions = this.additionalInput;
+      this.electronService.ipcRenderer.send('update-additional-extensions', this.additionalInput);
+      this.editAdditional = false;
+    } else {
+      this.modalService.openSnackbar(this.translate.instant('SETTINGS.extensionsInputError'));
```
whyboris, line 61: Nice touch! Showing error messages if something is wrong 😍

node/main-extract-async.ts
```diff
@@ -181,7 +181,8 @@ export function startFileSystemWatching(
       const ext = filePath.substring(filePath.lastIndexOf('.') + 1);
 
       // WARNING - dangerously ignores any path that includes `vha-` anywhere!!!
-      if (filePath.indexOf('vha-') !== -1 || acceptableFiles.indexOf(ext) === -1) {
+      if (filePath.indexOf('vha-') !== -1 ||
+          (acceptableFiles.indexOf(ext) === -1 && GLOBALS.additionalExtensions.indexOf(ext) === -1)) {
```
whyboris, line 185: no need to change, just a thought: `...` spread operator could fuse the two arrays before `.indexOf()` 🤔

whyboris:
I'll test it out sometime this week (🤞 hoping for today) 🤝 

The code looks good as far as I can tell 👍 

I am sure this is basically equivalent: another way to go would have been to just use the `wizard: WizardOptions` during `importFresh()` call - which would avoid having to create a new _IPC_ message 🤔 (again, doesn't matter, just thinking out loud here).

A really cool thing I like in JS now is the `...` _spread operator_ -- seems like an elegant way to combine arrays 👍

i18n/en.json
```diff
@@ -207,8 +207,11 @@
     "viewFolder": "View folder"
   },
   "SETTINGS": {
+    "applyAdditionalExtensions": "apply",
+    "additionalExtensions": "Support additional extensions",
     "autoGenerated": "Auto-generated tags",
     "buttonStyle": "Button style",
+    "cancelExtensionsEdit": "cancel",
```
whyboris, line 214: Thanks for alphabetizing too! 🙌 
I hope you're using a tool for that so you don't have to waste time 😉 
I like [sort lines](https://marketplace.visualstudio.com/items?itemName=Tyriar.sort-lines) and [sort JSON object](https://marketplace.visualstudio.com/items?itemName=richie5um2.vscode-sort-json) 🚀

orkomlosh:
_I am sure this is basically equivalent: another way to go would have been to just use the wizard: WizardOptions during importFresh() call - which would avoid having to create a new IPC message 🤔 (again, doesn't matter, just thinking out loud here)._

So, i thought about this, but i think what what i did also support if the user changes the extensions and than want to rescan a folder. without even go back to the wizard.
BTW I even managed to test this scenario and it worked. 

_A really cool thing I like in JS now is the ... spread operator -- seems like an elegant way to combine arrays 👍_
i agree it's much cleaner i'll add it.

whyboris:
Good thinking! I hadn't thought about users re-scanning a folder! 😁

I realize now, as we are adding more complexity to the app, it opens up more edge-cases we may need to handle 😓 . No need to change anything in this PR, but a thought ... if a user adds extensions while scanning a hub, but then removes the additional extensions they care about and re-scan, the _new_ settings get applied, removing the previously-added videos 🤔 

I think it's good-enough to just have what you've added, without tying the extra extensions to the `.vha` file 👌 I'll see how it looks and perhaps add an explanatory text behind an `i` information icon -- we'll see 🙆

orkomlosh:
from my tests it seems that if a file was already scanned it will remain on the hub that was created.
just a thought - maybe we can somehow add functionality to disable the edit extensions while scanning is still running.

whyboris:
The file will stay, but a re-scan goes through the same logic. If you include `.xyz` in the initial scan, those files appear. A re-scan after `.xyz` was removed from settings will not find the file and thus remove it from the hub (if I remember the logic correctly 🤔 ).

I don't have a strong opinion on disabling the extension while the scan is running. I think it's more work to catch a case that shouldn't happen often, and if it does, it doesn't break anything - just fails to find some files, right?

orkomlosh:
> The file will stay, but a re-scan goes through the same logic. If you include .xyz in the initial scan, those files appear. A re-scan after .xyz was removed from settings will not find the file and thus remove it from the hub (if I remember the logic correctly 🤔 ).
not sure, when i tested it, i added txt file to the supported it was added to the hub and then i removed the extension and rescan and it was still there.

whyboris:
It's possible that the current re-scan isn't working as expected. It's possible that you have to exit the current hub (close the app / switch to another hub) for the change to take place.

There's a `deleted` (optional) property on `ImageElement` that, when toggled, should hide the element from view.

A while back I added that after a re-scan, all the found files come back as a list `string[]` and if any of the current elements in `ImageElement[]` are no longer on that list, they should get the _deleted_ property set to `true` 👼 

I think that's the behavior I'd want -- remove any videos no longer found on the HD. In the past there were two buttons in the settings: (1) to only add new elements, and (2) to add new and remove old elements. I'm uneasy about decreasing options from people, but it seems most-expected that the app mirrors what you have on HD after a "rescan".

whyboris:
To be clear -- you an can keep your PR as is -- no need to worry much about how it behaves when we add/remove new file types to the list. 

The typical case is the only one I care about: if a user wants to add a new file type to a hub, they can do so in a clear, predictable manner. The rest are edge cases :trollface:

orkomlosh:
@whyboris I know that there's a lot of traffic and new issues this few last weeks and you have many things to handle. but i really want you to review this before it gets to the main branch 🤝

whyboris:
I'll do it after work today 🤝 in about 2-3 hours

whyboris:
🚀 super work! 
- works beautifully
- code looks great
🚢 ship it 😉 
If you think the individual commits are good to keep, `create a merge commit`, otherwise please `squash and merge` so the feature is added to history as a single commit 👍 

ps - delete behavior is as I expected - after removing the extension from the list and rescanning, the video gets removed, but unless `vhaNeedsSaving` gets updated (which is a bug on my end), the removal doesn't get saved into the `vha` file (I'll fix the bug myself there). *no need to change your PR - everything is working as I'd like* 😁

PR #536, opened 2020-09-08
Number of Videos in Folder sort

litelime:
A Sorting Option called "Number of Videos"
Sorts folders by how many videos are in them. 
In the case that a directory contains both single videos and folders, the single videos are considered "less than" any folder.


whyboris:
Thanks for the great addition! Sorry I've been away for a while (working on updating the official VHA website 😓 )
I'll try this out soon and merge it in 👍

src/app/pipes/sorting.pipe.ts
```diff
@@ -116,6 +118,32 @@ export class SortingPipe implements PipeTransform {
...
 
+    if (property === 'folderSize') {
+
+      // want non-folders to be considered "less than" a folder so give negative value by default.
+      var numCheck = /^\d+$/;
+      var xDisplay = -Infinity;
+      var yDisplay = -Infinity;
+
+      // numcheck because a folder will have num videos for this value, videos will have mb or gb characters for size
+      if(numCheck.test(x.fileSizeDisplay)){
```
whyboris, line 129: I wonder if there's a simpler way -- to check 
```js
if (x.cleanName === "*FOLDER*") { 
```
🤔 
I feel like co-opting a property rather than creating a new one for displaying folders may not have been the best move, but I'm still unsure 🤷 😅 

As an example, I check if something is a folder using the above approach: https://github.com/whyboris/Video-Hub-App/blob/main/src/app/components/views/thumbnail/thumbnail.component.html#L10

whyboris:
ps -- please ignore the *codeclimate* check (complaint) about this pull request 👌 -- I'm just trying out their service to see if it's useful 👍

whyboris (approved):
I tested the feature out -- it's brilliant 😍  I love this feature
Your code works as desired - I'm happy to merge in.
Consider the `x.cleanName === "*FOLDER*"` change -- or let me know and I'll merge in as is 👌

whyboris:
Thank you @litelime for the excellent feature addition 🙇 🤝 😁

PR #541, opened 2020-10-01
File System Async

### LINKED ISSUE #526: Current chokidar parameters need to be tweaked for performance
I've been experimenting with the new VHA today, and due to the current chokidar config, it's been crashing my systems kernels with an overload of filesystem operations! 😅 

I'm working on a better set of options, but here's some of the issues so far:

- The ignored list has changed, so now it only ignores files (vha-\*, images, etc.) after already adding a inotify watcher on it, and doesn't release the watch. This is a massive performance drop as it indexes my multiple vha-* dirs for no reason.

- Some of the params, like alwaysStat and awaitWrite are causing too much activity at one time, and would be better suited dealing with later in the process.

I'll update this with more findings as I experiment! 👍

### LINKED ISSUE #528: Manual thumbnail extraction is not async
When using the manual thumbnail extraction button, the verifying thumbnails part is synchronous, leading to a massive delay while the app is unresponsive! 😓 

We should do this with async methods, like the normal rescan version! 👍

whyboris:
Thank you @cal2195 for pointing this out 🙇 in #528 and at other times 😅 

Fixing things up 🧹 😁 👍 

- now we'll call `fs.stat()` at least one fewer times for each file during extraction 🎉 

Feedback welcome 🤝 
 - ps: I considered using `util.promisify()` but am uneasy about its magic 🤷 I'm open to using it if anyone recommends it strongly.

Closes #528
Closes #526


whyboris:
Hey @cal2195 -- thank you for the suggestion to make more of my app _async_ 🙇 

If you happen to have time, could you glance over the changes just to see if anything jumps out as a mistake? I've not dealt with async stuff much 😅 

Most of the changes are white-space changes, so look with the `?w=1` in the url, like this:
https://github.com/whyboris/Video-Hub-App/pull/541/files?diff=split&w=1

whyboris:
@cal2195 -- I'm merging as is but if you have the time to glance over the changes I'll appreciate it 🙇 

No rush -- I think I'm at least 1 month away from releasing version `3.0.0` 😅

PR #542, opened 2020-10-02
Updated the main README.md and icon README.md 

Shahzeb-Kazmi:
Updated README.md to enhance readability and understandability


whyboris:
Thank you for the contribution!

README.md
```diff
@@ -50,12 +50,14 @@ Main dependencies in use:
 | Angular-CLI      | v10.0.0 | June 2020   |                                                   |
 | Electron         |  v9.0.0 | June 2020   | (internally uses Node `v12.14.1` and Chromium 83) |
 | Electron Builder | v22.7.0 | June 2020   |                                                   |
-
+|
```
whyboris, line 53: This might be accidental? It looks like this:

image

README.md
```diff
@@ -4,17 +4,17 @@
 
 Read in [Português (Brasileiro)](https://github.com/whyboris/Video-Hub-App/blob/master/README.br.md)
 
-## Download now
+## Download Now
 
 This software is available for $3.50 through [videohubapp.com](https://videohubapp.com/download.html)
 
-$3.50 of every sale goes to the [_cost-effective_](https://www.givewell.org/charities/top-charities) charity [Against Malaria Foundation](https://www.againstmalaria.com/).
+$3.50 of every sale goes to the [_cost-effective_](https://www.givewell.org/charities/top-charities) charity of [Against Malaria Foundation](https://www.againstmalaria.com/).
```
whyboris, line 11: I think the "of" is not needed grammatically 🤔 

Example: "I drove across the bridge Golden Gate" or "A part of my spending money goes to the musical band Pink Floyd"

Thoughts?

whyboris (commented):
Thank you for the improvements! 

I think there are 2 changes that are not needed - let me know your thoughts 🤝

Shahzeb-Kazmi:
Hello,

The first one seems like an unintentional error however the in the second
one including 'of' suggests the amount in question is a part of the sale.
If 'of' doesn't seem right, 'from' is a more clear choice for this instance.

I hope this helps.

Best,
Shahzeb

whyboris:
How about a re-wording of the whole thing?
```md
$3.50 of every sale goes to the [Against Malaria Foundation](https://www.againstmalaria.com/), a [_cost-effective_](https://www.givewell.org/charities/top-charities) charity rated highly by [GiveWell](https://www.givewell.org).
```
Also, are you suggesting `$3.50 from` rather than `$3.50 of` ? Either one is fine -- seems like the same meaning to me.

Shahzeb-Kazmi:
Rewriting the entire text isn't necessary, it is perfectly clear.

"3.50 of" and "3.50 from" do entail the same meaning, but "from" is more
direct and explicit in comparison to "of" hence a better option, in my
opinion.

Best,
Shahzeb

whyboris:
Sounds good. I'm still urging reverting your change from including `of` before AMF -- and keeping the sentence as it was before your change.

whyboris:
@Shahzeb-Kazmi thank you for the proposed changes. I decided to resolve the conflicts myself and revert one of your proposed changes to how it was before.

Thank you for the contribution 🤝

PR #548, opened 2020-10-08
Turkish Translation

bereker:
I translated comparing en.json and tr.json realized that there were 4 different lines that tr.json didn't have I added and translated them accordingly.
I improved on the Turkish translation overall making it easier to comprehend. 
And also fixed a small typo on en.json


whyboris (approved):
Thank you so much!!!

PR #549, opened 2020-10-09
Use birthtimeMs instead of ctimeMs as file created time

CrendKing:
I'm sorry I made a mistake in https://github.com/whyboris/Video-Hub-App/commit/3e58686fc2015efe13adb76f66ba10eefd762eca that I used `ctimeMs` as file created time. According to [the document](https://nodejs.org/api/fs.html#fs_stats_birthtimems), `birthtimeNs` is the correct property name. `ctime` stands for "change time", just like the [Linux equivalent](https://linux.die.net/man/2/stat).

I tested the change. Should be OK now.

### DIFF
--- a/interfaces/final-object.interface.ts
+++ b/interfaces/final-object.interface.ts
@@ -33,7 +33,7 @@ export interface ImageElement {
   height: AllowedScreenshotHeight; // height of the video (px)
   inputSource: number;           // corresponding to `inputDirs`
   mtime: number;                 // file modification time
-  ctime: number;                 // file creation time
+  birthtime: number;             // file creation time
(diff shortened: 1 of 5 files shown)


whyboris (commented):
Thank you! Looks good 👍 

Since the previous change hasn't yet been released into "production", we can change the interface 👌

whyboris:
I tested locally -- works great. The value is a timestamp so it sorts as expected 😁

CrendKing:
Thanks for fast merge!

When I test this change, I needed to completely rebuild the whole vha profile, because the property name is changed. I wonder if we should introduce some mechanism to detect and refill missing properties in vha file during rescan/monitoring, so that in the future if you or anyone introduce new property, users don't have to rebuild the profile again and again.

whyboris:
It seems like I could add that functionality whenever needed; I don't think I need to add it in advance of a feature 👍

Issue #352, opened 2020-01-28
Folder flattening


danjool:
Potentially valuable to seem to streamline some of their media library flattening the contents of a chosen folder -- moving all the media contents of that target folder's subfolders to the target folder.

https://github.com/whyboris/Video-Hub-App/pull/351
(ONLY TESTED IN WIN10)
Demonstrates this by adding a right click context menu to the folder, replicating the context menu for media files but only supplying a Flatten command.  If this feature were approved, implemented this way, I'd expect a few other folder commands would make sense as well.

Two design questions:
Q1: In a typical filesystem context I would take that to mean all subfolder contents bubble up to the target folder, and the now useless subfolders are removed.  In the context of VHA, that seems like over reach, in that the user would be deleting things they can't see, the non media files and the subfolders without media files.  Does that seem right?

Q2:  I'm torn on how the user should be able to engage this action.  The way I implemented it is that all subfolders of the target folder a user clicks on bubble up, the target folder stays where it is.  An alternative, plausible user expectation that the target folder itself would also bubble its contents up to the current folder.

whyboris:
Thank you for the discussion and for the PR #351 🙌 

I'll give the code you propose a try and see how it feels. I agree with you that VHA restructuring all the folders would be an "over reach". 

I'm unsure yet about the functionality: I do like that VHA provides a faithful representation of what the folder structure is, but this flattening is definitely useful in some situations. My immediate concern is all the code complexity and questions that need to be answered with code (for example, what happens if users add more files into the subfolders and re-scan, etc).  

I'll comment again after I try it out 👍

danjool:
My expectation would be that if I

1. Flatten contents out of some subfolders

cal2195:
Just leaving my thoughts here: 😄 

I appreciate the work here, although I'm not sure it fits within VHA - this seems more like an idea for a script, eg.
```
for file in subfolderfiles:
    mv file .
```

Even deleting files in VHA only moves them to the trash - there's a lot of headache dealing with filesystem errors that'd I'd rather not deal with! 😓 Like, what happens if two files are now the same name, trying to move to the same folder..?

On the other hand - flattening a file system is _really_ useful sometimes, but I think we should implement it in the way things like `ranger` implement it - that is, don't change the file system, but only change the view in VHA.

[...]

PR #563, opened 2020-10-21
rename folder

### LINKED ISSUE #511: Rename folder
Would it be possible to add rename option into context menu for folders, just like it is possible to rename files/videos (on disk)? Would help me to organize my data/folders directly in VHA without a need to re-scan over and over again. Thank you.

Btw, are the auto tags generated from folder names too? (use-full for me)

shubham13695:
Closes #511

### DIFF
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -422,8 +422,8 @@ export class HomeComponent implements OnInit, AfterViewInit {
     this.electronService.ipcRenderer.on(
       'rename-folder-response', (
           event,
-          index: number,
           success: boolean,
+          index: number,
           renameTo: string,
           sourceFolder: string,
           errMsg?: string
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -1890,7 +1890,9 @@ export class HomeComponent implements OnInit, AfterViewInit {
     this.currentRightClickedItem = item;
     this.rightClickShowing = true;
   }
-
+  ok(){
+    console.log("ok")
+  }
   /**
    * When in double-click mode and a video is clicked - `currentClickedItemName` updated
    * @param item
--- a/src/app/components/home.component.ts
+++ b/src/app/components/home.component.ts
@@ -417,28 +417,28 @@ export class HomeComponent implements OnInit, AfterViewInit {
           });
           this.renameFileResponseBehaviorSubject.next(undefined); // allways remove right away
 
-      });
+    });
 
-      this.electronService.ipcRenderer.on(
-        'rename-folder-response', (
-            event,
-            index: number,
-            success: boolean,
-            renameTo: string,
-            sourceFolder: string,
-            errMsg?: string
-          ) => {
+    this.electronService.ipcRenderer.on(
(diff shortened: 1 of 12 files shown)


whyboris:
Really cool 🎉 Thank you for the PR 🤝 I'll try it out later today 🏃‍♂️

whyboris (commented):
After a successful folder rename, do the videos in that folder remain playable?

I feel like unless we update all the paths for the videos in that folder, the videos in that folder will never be playable again (via clicking in the app) 🤔

src/app/components/home.component.ts
```diff
@@ -1890,7 +1914,9 @@ export class HomeComponent implements OnInit, AfterViewInit {
     this.currentRightClickedItem = item;
     this.rightClickShowing = true;
   }
-
+  ok(){
```
whyboris, line 1917: WIP [spandrel](https://en.wikipedia.org/wiki/Spandrel_(biology))? 🙂

whyboris:
@shubham13695 -- please let me know if you'd like to address the concern I raised or you'd rather I finish up the feature after merging 👌 

At the moment I'm really tempted to get version 3.0.0 released - and this feature could be added after (so there's no rush now). 

Thank you again for your contribution 🙇

shubham13695:
@whyboris  yes I address it that after changing the folder on physical directories we need to update that change in our app. I m figuring out how to do it as `rightClickElement` index return -1 . if you have any idea please share it with me.

whyboris:
I think this might only need a method in the `ImageElementService` to update all the objects in the `imageElements` array 🤔 (replacing the `partialPath` with the updated version of it).

So once the `rename-folder-response` comes back to _Angular_ with a success, you would call that method 🤔

shubham13695:
hi @whyboris 
I don't know what wrong with `imageElementService.imageElements` like before I update the file partial path it already has updated the folder with an older one same thing happening with the video file also. 
if i update the same video file twice it create one more file.

whyboris:
Cool! I checked out your branch -- somehow it creates the new elements (I'll look through your code again -- I didn't realize it did that somehow).

All you'll need to do is mark every element in the old folder path as `deleted: true`:

whyboris:
btw -- if you happen to know -- could you point out how the new elements get created with your code? I'm looking all over your new code and I don't see anything that adds the videos from the new renamed folder into the app gallery - yet the videos do show up 🤯 :trollface: 

I don't understand how that happens from your code -- I'm glad that it does, but could you let me know if you happen to know? 😊 😅

Kendra666 (approved):

whyboris:
I'm hoping to release version `3.0.0` this week. Because of limited times, and prioritizing the release, I chose not to include this feature in the initial `3.0.0` version. 

I look forward to merging it in after a successful release of `3.0.0` 🤝

whyboris:
I'm eager to release version `3.3.0` so this PR will wait for another release 🙇‍♂️ 

I appreciate the contribution and will try to get it out with a future release 🙇‍♂️

whyboris:
Related issue #352 👍

PR #585, opened 2020-11-18
Copy meta from recently deleted to new if matching

### LINKED ISSUE #584: Add any tags & meta from "deleted" file to "new" file
If a user renames a file using the OS, or moves the file from one folder to another, the app will remove the no-longer-there file from its gallery, and if re-scanning (or "watching" is enabled), the app will add the "new" file into its gallery.

Since removed files exist in the app memory until the app is restarted or user switches to another hub, we can simply perform a check upon adding a new file: check whether any of the files marked as `deleted: true` match the hash, in which case copy over their meta into the "new" file.

Should be easy to do 😎

whyboris:
Closes #584

Unsure whether I need `\\` before path or not 😓


src/app/components/home.component.ts
```diff
@@ -578,12 +578,13 @@ export class HomeComponent implements OnInit, AfterViewInit {
...
 
     // When `watch` folder and `chokidar` detects a file was deleted (can happen when renamed too!)
+    // mark the element in `imageElements[]` as `deleted`
     this.electronService.ipcRenderer.on('single-file-deleted', (event, sourceIndex: number, partialPath: string) => {
       this.imageElementService.imageElements
         .filter((element: ImageElement) => { return element.inputSource == sourceIndex })
         // notice the loosey-goosey comparison! this is because number  ^^  string comparison happening here!
         .forEach((element: ImageElement) => {
-          if (('\\' + partialPath) === path.join(element.partialPath, element.fileName)) {
+          if (partialPath === path.join(element.partialPath, element.fileName)) {
```
whyboris, line 587: I'm unsure why I had `\\` before -- I think without it, the functionality didn't work.

But when trying today, it did.

I think my `.vha2` files might have a different path strings. I need to make sure somehow they are normalized.

Perhaps `path.normalize()` 🤔

whyboris: Just going to leave as an `||` for both scenarios 🤷‍♂️ 
Works well-enough 👌

Issue #562, opened 2020-10-20
Add star rating on right of top ribbon


whyboris:
Should be pretty easy. After having clicked a video, the folder path and file name show up on the "top ribbon". On the right side of the ribbon there is space for 5 stars. Would be the easiest way to rate videos ⭐ ⭐ ⭐ ⭐ ⭐

whyboris:
Thank you @qqweng for the PR 🤝 #586

I'm wondering whether it's better to relegate the star rating to the bottom tray 🤔 or just have both options 😓 

I'm unsure about the star-rating at the top:
- what happens when scanning is happening and the donut progress bar is showing?
- since we'll have the _details tray_ - won't it be overkill to have the star rating in the top-right as well?
- at the very least, we'll want a button to hide the star rating from the top ribbon then? 🤔 

@qqweng -- let me know your thoughts on this 🤝

RumaboyLotti:
Hey would like to take a gander at this for my first issue I am assuming it is still open but let me know if that changes

whyboris:
@RumaboyLotti -- thank you for offering your support. 

I've not thought about this issue in a while, and currently I'm working on house stuff (with intent to get back to VHA development this winter) 😅 

Now that we have the "details" tab at the bottom - with the star-rating already present there -- maybe star rating at the top isn't even needed. 

If we do put it at the top, this is a challenging feature to add (from the UI - user interface - perspective). I am unsure why I added `good first issue` 😅 

From the technical side it seems pretty straightforward, but resolving the UI questions in the above comment takes some thought.

If you have ideas to address the UI issues - could you post a screenshot? (A simple screenshot of the app with MS Paint--or similar low-quality mock up--is enough). 

If this is a feature you particularly want - please share a screenshot of how you think it could work; if you're just eager to try something out, please consider another _Issue_ ❤️

whyboris:
Since the _Details view_ contains the star-rating section, I don't see a reason to add stars to the ribbon. Closing the issue 🙇‍♂️

PR #586, opened 2020-11-20
Adding feature star rating to top ribbon #562

qqweng:
This is my first time contributing. Please let me know if there are any issues or concerns, and I will try my best to fix :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "trash": "6.1.1"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "0.1000.0",
+    "@angular-devkit/build-angular": "^0.1100.0",
     "@angular/animations": "10.0.9",
     "@angular/cdk": "10.1.3",
     "@angular/cli": "10.0.0",
--- a/src/app/components/top/top.component.html
+++ b/src/app/components/top/top.component.html
@@ -65,7 +65,9 @@
   </div>
 
   <app-icon class="icon" [icon]="'icon-video-blank'" (click)="openInExplorer()" style="margin-left: 0; "></app-icon>
-  <div style="margin-left: 23px; width: 80%;">
+  <div 
+    style="margin-left: 23px; width: 80%;"
+  >
     <span
       *ngFor="let item of fileNameArray"
       (click)="fileWordClicked(item)"
--- a/src/app/components/top/top.component.html
+++ b/src/app/components/top/top.component.html
@@ -18,54 +18,16 @@
   class="file-container"
   [ngClass]="{ 'dark-mode-override': darkMode }"
 >
-  <!-- Star rating --> 
-  <div 
-    class="starRating"
-  >
-    <button
-      (click)="setStarRating(1.5)"
-      class="star-rating"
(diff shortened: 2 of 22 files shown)


whyboris:
Really cool! Thank you 🙇 

I'll test it out today or this weekend and let you know if anything needs changing -- by looking at the code, looks good!

src/app/components/top/top.component.ts
```diff
@@ -52,4 +76,16 @@ export class TopComponent {
     this.onOpenInExplorer.emit(true);
   }
 
+  setStarRating(rating: StarRating): void {
```
whyboris, line 79: This is duplicating `setStarRating` in the `meta-component` - so it's a good time to create a new component that will have this functionality 👍 and we can then remove this method from both here and `meta-component` 👍

src/app/components/top/top.component.html
```diff
@@ -18,12 +18,62 @@
...
-    (click)="fileWordClicked(item)"
-    class="file-word"
+  <!-- Star rating --> 
+  <div 
+    class="starRating"
   >
-    {{ item }}
-  </span>
-</div>
+    <button
```
whyboris, line 25: This section is duplicating the same code as happens in the `meta-component` so it would be best to just create a new component which will have this chunk in its template, and the `setStarRating` method in its `component.ts` file 👍

package.json
```diff
@@ -48,7 +48,7 @@
     "trash": "6.1.1"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "0.1000.0",
+    "@angular-devkit/build-angular": "^0.1100.0",
```
whyboris, line 51: I don't mind too much, but it's generally better not to change versions and keep a PR focused on one thing. 

We could update to Angular 11 in a separate PR 🤝

whyboris (changes requested):
Really cool! It works 😁 

I am recommending that the star rating template, css, and component code all becomes part of its own _component_ - so we can just re-use it in both places 👍 

Please let me know if you would like any guidance or help -- any time! 🤝

whyboris:
@qqweng -- I'm going to be adding a "Details" tab to the bottom tray: #587 

It will let users change the star rating of the last-clicked video. I wonder if it still makes sense to add a star rating system to the top bar as well 🤔 

Please let me know what you think 🤝

qqweng:
@whyboris 

If there's going to be a details tab on the bottom tray, I feel that would definitely be better than having the star rating component on the top component. Especially since users would be able to edit and makes changes to the video details a lot easier that way. 👍 

I did make changes to separate the star rating component into its own component. It may help with future changes (or not 😅 ). 

Feel free to discard the changes (I know this issue has been open for a long time so you probably considered other ideas as well). 

I learned a lot, and I'm glad I found this project (first time working with Angular and Electron). I will continue to look out for other issues 😄

whyboris:
Thank you again @qqweng for the contribution 🙇‍♂️ 

I'm sorry I'm choosing not to merge this change in as the _Details view_ is sufficient for rating videos 😅

PR #587, opened 2020-11-21
Details tray for last clicked video

whyboris:
Easiest way to edit details for a video 🚀 
When you have the _tray_ open with the _Details_ tab showing, the last-clicked video will be displayed along with all its meta details (and you can easily rename the file from there too 😎 )
Will be an easy place to add ratings to a video 👍 

🤔 Unsure if we'll still want #562 😅


whyboris:
minor bug remains: changing the number of stars updates the video, but does not update the "Details View" if it's open along with the "Details Tray" 🤷 

Not a problem -- I'm OK with not fixing it at least for now.

PR #593, opened 2020-11-28
Angular 11

whyboris:
Everything seems to work 😅 

Updated everything I could except for `electron` which is still stuck: https://github.com/electron/electron/issues/23757 😢


whyboris:
Confirming it builds on Windows 10 and the extraction functionality works 🚀

PR #611, opened 2020-12-18
Transcode server for remote

whyboris:
Huge thank you to @cal2195 for his code 🙇 

This code comes from his `transcoding-video` [branch](https://github.com/whyboris/Video-Hub-App/compare/main...cal2195:transcoding-video) 🥇 

This is meant to be released in tandem with https://github.com/whyboris/Video-Hub-App-remote/pull/9 🤝 

### ~Help wanted~

_update: Thank you @cal2195 for the `on-finished` suggestion -- I think it fixed the below problem_

When running this branch in tandem with the _Remote_ - after a little bit (playing the video 10-20 times) the server fails:
```
x264 [error]: malloc of size 2840448 failed
Video encoding failed
```
I messed on my PC with the _pagefile_ settings, so it could be me. But I also wonder if, since for every HTTP request that comes to this server, we're `spawn`ing a new `ffmpeg` process, and perhaps never killing it, something goes wrong? 😅 

### Help still wanted:

Also, I set `'-frag_duration', '15'` to test a different value, but originally it was `3600` ... I'm unsure about what it's meant to do - the unit is seconds, and 3600 is 1 hour which seems awfully long (though I still don't understand what the flag does) 😅 

Documentation is conflicting (see comment in discussion below).


cal2195:
Sounds like you're correct - we need something like this to catch the connection closed event, and then `.kill()` the ffmpeg process, possibly with the correct signal: ⚠️ 

https://github.com/jshttp/on-finished

cal2195:
Also note, for MOV/MP4/ISMV (Smooth Streaming) muxer, `-frag_duration` is in _micro_seconds, not seconds! 😄 

https://ffmpeg.org/ffmpeg-formats.html#Options-9

whyboris:
@cal2195 unsure if you were kidding 🤔 
The [FFmpeg documentation](https://ffmpeg.org/ffmpeg-formats.html) seems to say _seconds_ 😅 
> `frag_duration` _duration_
>     Set the length in seconds of fragments within segments (fractional value can be set).

whyboris:
Thanks @cal2195 -- I think the `on-finished` did the trick 🤞 https://github.com/whyboris/Video-Hub-App/pull/611/commits/44ead957aa1be366c7e5e8b006c4bcd84f4f07a4

cal2195:
> @cal2195 the [FFmpeg documentation](https://ffmpeg.org/ffmpeg-formats.html) seems confusing 😅
> 
> > `frag_duration` _duration_
> > Set the length in seconds of fragments within segments (fractional value can be set).
> 
> > `-frag_duration` _duration_
> > Create fragments that are _duration_ microseconds long.
> 
> :trollface:
> 
> Also, _microseconds_ !?!! 😱 that's [1,000,000 microseconds in 1 second](https://www.convertunits.com/from/second/to/microsecond) ... that's a _million_ :trollface: ... they must have messed up in the documentation, right? Especially that they allow _fractional value_ ... right? 😖

The one that is in seconds is for the dash muxer, each muxer had it's own options. :) Only the one in seconds allows fractional values.

We want the mp4 muxer, which is in microseconds! 😄

whyboris:
I don't know what's a sensible values (as measured in _seconds_ is). Is there a way to tell whether the value affects anything?

I can't see any difference whether it was `3600` as originally, or when it's `15` :trollface:  - what would I expect to happen when I change it? Perhaps it gets ignored if it's too short to matter, and any effect is only noticeable at values above 1 million (1second)?

Sorry to ask about questions that I should be able to research on my own -- but the documentation is so terse 😕 I glanced around online (not too long, just a quick search) and found nothing of value to explain.

Or if it's easier, please let me know how you figured this stuff out 🙇

node/server.ts
```diff
@@ -94,6 +102,38 @@ function startTheServer(pathToServe: string, port: number): void {
...
+    const file = req.query.file || '';
+    // see https://trac.ffmpeg.org/wiki/Encode/H.264#a2.Chooseapreset for more options
+    const ffmpeg = spawn(ffmpegPath, [
+      '-ss', seekTime,
+      '-i', file,
+      '-f', 'mp4',
+      '-crf', '17',
+      '-preset', 'ultrafast',
+      '-movflags', 'frag_keyframe+empty_moov+faststart',
+      '-frag_duration', '15',
```
cal2195, line 117: You can probably remove this - I think it might have a sensible default anyway. We can change it later.

The idea with fragmenting, is that you can seek and stream much easier, without having to download the entire mp4 file before playback can begin. Each fragment contains metadata about the video, rather than all the metadata being store at the end of the file!

cal2195: I can look into finding the best options, but if the code works as is, may as well use that for now!

cal2195 (approved):

node/server.ts
whyboris: Thanks Cal for looking at this again. I've paused this feature for a while because I had at least two days of trying hard with no success: getting the video to play in Chrome ✅ no problem 👌 but getting Safari on iOS (on my phone) to play anything was a no-go 😓 which is the sole purpose of this feature.

I'll come back to this with renewed energy once I move in to the new house (these few months have been busy).

😅 also I've been working on an image-centric analogous-to-VHA app this past month -- I want to finish it first before returning to this headache :trollface:

whyboris:
I intend to glance around to see whether sending video to the phone is easier in 2026 - after release of VHA `3.3.0` 👍

PR #612, opened 2020-12-18
Add "times played" slider filter

litelime:
In order videos with 0 plays to show, min value needed to be set to -1. 
So display pipe adds 1 to value to keep displayed label value 0+.
The effect is that left slider is inclusive, and right slider is exclusive to the actual number of times played.
I think it is still pretty straightforward to use though.

One thing I noticed is the view does not immediately update when the times played value goes outside the bounds of the filter. 
EX: With the stars filter, if I am filtering for only 5 star videos, and I rate a video less than 5, it will immediately leave the view, would like to get that same behaviour here


whyboris:
Super cool! Thank you for the addition 😁  -- I'll try to review and comment this weekend. From glancing at the code so far -- looks good and clean 🥇

whyboris:
Sorry didn't get to it this weekend 😅 
Will try today/tomorrow 🤝

whyboris:
Thank you @litelime 🎉 stellar work 🤩 
Sorry took so long to merge 😓 have been busy with non-coding life stuff 👌 (nothing bad, just time consuming).
But I've still not released `3.1.0` so this will be in the release whenever that happens 👌 😅 👌

Issue #623, opened 2021-01-12
Extract FPS for each video


whyboris:
Seems like it could be useful 🤷‍♂️ 

If it's easy enough to grab it from the JSON object from FFprobe 🤔 

- Add as an extra field to `ImageElement`
- In _Details view_ display after the size, e.g. `| 18mb | 30fps | YYYY`
- Probably best to round to nearest integer (I hate `29.97` fps :trollface:  - so annoying) - so `24`, `30`, `60` etc
- Add _sort by_ filter for convenience 👍 

Open to a discussion 🤝

jleckron:
Hi Boris, if this has not already been implemented, I would like to take a stab at it. Could you give me some guidance as to test this is working?

whyboris:
The metadata extraction happens in this method: `extractMetadataAsync`
And we parse the JSON coming from `ffprobe` which should contain the FPS in there somewhere 😅 
https://github.com/whyboris/Video-Hub-App/blob/main/node/main-support.ts#L415

The FPS is likely going to look like this: https://stackoverflow.com/questions/7708373/get-ffmpeg-information-in-friendly-way
`"r_frame_rate": "30000/1001",` - "Real base framerate of the stream."

Store the `Math.round(fsp)` in the `ImageElement` and you're good to go 🚀 

Let's store it as `fps` -- just add an entry to the _Interface_
https://github.com/whyboris/Video-Hub-App/blob/main/interfaces/final-object.interface.ts#L27

Just add the `{{ video.fps }}` in this file: https://github.com/whyboris/Video-Hub-App/blob/main/src/app/components/meta/meta.component.html#L35 to see it working 😎

jleckron:
Thank you for the help. I was having trouble finding the file where the fps was displayed, so that clears a lot up. How would I go about running a test environment of the application? I am probably doing this incorrectly, but when I open src/index.html in a browser, I get stuck on the loading animation.

whyboris:
I didn't know videos can be variable FPS 😅 
I'm a little unsure. Are both available from `ffprobe` and you have to choose, or would one of them need to be computed manually? I'd opt for whatever `ffprobe` provides 🤷 I suspect the two are identical for 99.9% of the cases anyway 🤔 
Pick whichever of the two is most appropriate 🤝

jleckron:
Ok sounds good. I am by no means an expert on video either, so I was just an unsure as you 😅
[...]

PR #635, opened 2021-02-02
Add fps

jleckron:
Hi Boris,

I added functionality to see the real frame rate for each video in detailed mode 😉

I also added functionality to use the sidebar to sort by fps. In doing so, I added a couple new key-value pairs in `en.json`. According to `i18n/README.md`, hopefully it is simple to expand my changes for all other supported languages. 

Please let me know of any changes you would like to see.

Thank you for the support!

Issue #623


interfaces/final-object.interface.ts
```diff
@@ -39,6 +39,7 @@ export interface ImageElement {
   stars: StarRating;             // star rating 0 = n/a, otherwise 1, 2, 3
   timesPlayed: number;           // number of times the file has been launched by VHA
   width: number;                 // width of the video (px)
+  frameRate: number;            //  base frame rate of the video in fps
```
whyboris, line 42: I don't have a strong preference, but perhaps `fps` is better - as it's shorter and is already a well-known acronym 🤔

interfaces/final-object.interface.ts
```diff
@@ -56,6 +57,7 @@ export interface ImageElement {
   resBucket: number;             // the resolution category the video falls into (for faster sorting)
   resolution: ResolutionString;  // e.g. `720`, `1080`, `SD`, `HD`, etc
   selected?: boolean;            // for batch-tagging of videos
+  frameRateDisplay?: string;      // displayed as XXfps
```
whyboris, line 60: I don't think this field is needed - as we can just append (the text) "fps" in the _template_ (the `.html` file) 👍

node/main-support.ts
```diff
@@ -133,6 +133,19 @@ function getDurationDisplay(numOfSec: number): string {
   }
 }
 
+/**
+ * Generate frame rate display formatted as XXfps
+ * @param frameRate
+ */
+ function getFrameRateDisplay(frameRate: number): string {
```
whyboris, line 140: I think this is better done in the _template_ (`.html` file) 👍 

The reason some other values are stored in the `ImageElement` object rather than live-computed is because their computation is much more computationally heavy, while this one is just appending text 👍

src/app/components/meta/meta.component.html
```diff
@@ -34,6 +34,8 @@
 
     | {{ video.fileSizeDisplay }}
 
+    | {{ video.frameRateDisplay }}
```
whyboris, line 37: Here is where we can add the "fps" text 🤝

whyboris (changes requested):
Awesome work! 

Thank you for the comment about the _real_ vs _average_ FPS 👍 

I'm requesting just a minor change: moving the "fps" text to the template 👍

whyboris:
Thank you @jleckron -- updated code looks great! I'll test it before merging this weekend 🤝

whyboris:
I tested -- works beautifully. I'll update the translations for other languages and add the button to the "Sorting options" list ✅ 

Excellent work - thank you very much 🙇 🤝 🎖️

whyboris:
I made a few quick changes -- making the FPS search toggleable in the settings 🤝 
#636 🙇

PR #647, opened 2021-03-10
Adding text-shadow to time/rez classes

TheTacoScott:
Makes the time and rez metadata display exceedingly more readable.

### DIFF
--- a/src/app/components/views/time-and-rez.scss
+++ b/src/app/components/views/time-and-rez.scss
@@ -3,6 +3,15 @@
   color: rgba(255, 255, 255, 0.8);
   font-size: 10px;
   position: absolute;
+    text-shadow:
+    -1px  -1px  3px #000,
+    0px   -1px  3px #000,
+    1px   -1px  3px #000,
+    1px    0px  3px #000,
+    1px    1px  3px #000,
+    0px    1px  3px #000,
+    -1px   1px  3px #000,
+    -1px   0px  3px #000;
   z-index: 2;
 }
 


whyboris (commented):
Thank you for the PR 😁 

I do dislike that there are times when the time & resolution are hard to see (when the background of the screenshot matches too much with the text). But I'm unsure how to best solve this problem.

I am rather against `text-shadow` as it is "expensive" performance-wise. Especially if it's such a complicated text shadow. 

A semi-transparent background on the text element would be much less performance-heavy - but it doesn't look very good. Adding a border-radius is also "expensive" but would likely be needed to make it look better. 

I'm unsure about how to _visually_ address the problem I've had in the app since version 1 😅 

I'm open to a discussion. Feel free to share screenshots 🤝

TheTacoScott:
Good example of what it looks like with a white background on left half of thumbnail.
image

The (Tags: 2) is my addition on my local branch.

TheTacoScott:
More generic video
image

TheTacoScott:
Background where it's addition doesn't matter much:
image

TheTacoScott:
Example where it's impossible to see without text-shadow (or something like it):
image

Same thing with text shadow:
image

TheTacoScott:
While zoomed in it looks a bit blurry, but in natural size it looks clean and sane.
image

TheTacoScott:
Good example of when it is basically unreadable without shadow
image

TheTacoScott:
Much easier with shadow:
image

TheTacoScott:
Unreadable: 
image

Readable:
image

TheTacoScott:
Those are the readability improvement screenshots. I think there is a case to improve the readability. But you bring up a very good and valid point: what is the right way to do it?

Outside stroke (aka: text shadow 8 axis) makes it pop a bit, but it's not a free operation in css.
Span/Div Opacity changes make it all look a bit too boxy for my taste.

here is an example of adding some padding/and setting a background color and some opacity:
image

TheTacoScott:
Fairly easy to get a computationally cheap CSS stroke effect with text-stroke (interior stroke) but then the font size has to be bumped up a lot;
image

TheTacoScott:
mix-blend-mode might have some options:
difference:
image
image

whyboris:
Thank you for the screenshots and ways to address the current problem. I'm a bit busy this coming month - but I'll return to this topic when I find the time again 🤝

TheTacoScott:
No rush. In fact I think the most proper way to solve this might be the same way you display the "total time of videos in folder" in folder view. It's a simple border radius box. While that is less 'cool' than fancy opacity/text shadows/etc it would be more in line with the existing visual language.

whyboris:
Thank you @TheTacoScott for the PR 🙇‍♂️ 

I'm releasing version `3.3.0` with a related change I made - making the time and resolution easily readable 👍 

Closing as this change is no longer needed ❤️

PR #680, opened 2021-07-18
space bar plays random video

whyboris:
🚀 👨‍🎤 😎 

new settings button 😉


whyboris:
🤔 I still like this feature - I think I'll add it 😎

Issue #691, opened 2021-08-12
CMD+H and CMD+M not working on Mac


dia3olik:
Hey Boris!

I just discovered these two shortcuts don't work with VHA3.

Every app systemwide usually adheres to at least these 2 shortcuts to HIDE and MINIMIZE the App.

Is there a chance to see these implemented like other Mac Apps?

Thanks!!!!

whyboris:
Thank you @flpcury for your fix PR 🙇 it works great 🎉
Thanks for your patience - took me a while to get to review it 🤝

PR #692, opened 2021-08-13
Add more macOS menu/shortcuts

flpcury:
Requested feature (#691)
Added a "Hide `app.name`" menu item and shortcut.
Added "Window" and "View" labels (submenus).
The `window` menu role comes with not only "Minimize", but also with zoom and tiling functionality.
But if, for example, you tile the window, it goes into sort of a Full Screen mode (but occupying half the screen). That's normal behavior in macOS, but there's no way to leave Full Screen mode without having the "Exit Full Screen" item/shortcut on the menu, so I added a "View" label with a "Enter/Exit Full Screen" item as well.

### DIFF
--- a/main.ts
+++ b/main.ts
@@ -119,9 +119,16 @@ function createWindow() {
         ]
       },
       {
+        label: "View",
+        submenu: [
+          { role: "togglefullscreen" },
+        ]
+      },
+      {
         label: "Window",
         role: 'windowMenu',
-      }]));
+      },
+    ]));
   }
 
   // Create the browser window.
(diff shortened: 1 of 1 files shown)


whyboris:
Thank you so much! 

I'll try to find time this weekend to try it out. I am 99.9% sure it works, but still want to run it once on my machine 🤝 🙇

whyboris (approved):
Finally got to reviewing 😅 thank you for your patience.

I'm LOVING this 🤩 -- thank you 🙇

Issue #681, opened 2021-07-19
Scroll to top when switching settings tabs


whyboris:
Currently the app does not scroll to top when switching between settings tabs. This is mildly uncomfortable. Let's make it scroll to top when switching to another tab 🤝

dVp007:
Can you please add a gif if possible ?

whyboris:
To replicate:

1. Open settings menu
2. Scroll to the bottom of the settings (until you see "check for new version")
3. Click on 1st tab "Search settings"
4. Notice that "Filter & sorting options" is showing (and the scroll-bar is mid-way through the page) instead of the top ("Search & filter settings")

Please let me know if you'd like a better description 🤝

dVp007:
https://github.com/whyboris/Video-Hub-App/pull/700

Please take a look at this pr. :)

Adityamaurya123uma:
@whyboris Is this issue still open?

Adityamaurya123uma:
@whyboris I am facing issues in starting this project locally
![image](https://github.com/whyboris/Video-Hub-App/assets/63842678/cd010248-f36e-4892-b8ca-85203434c513)
Locally I downgraded the TS version to 5.1.6, but the error still remains.

whyboris:
Sorry missed this message somehow 😓 

Please use `npm install --legacy-peer-deps` when installing dependencies (see [Readme](https://github.com/whyboris/Video-Hub-App#development))

Please let me know if this doesn't fix the problem 🤝

whyboris:
Thank you @sam9525 for the fix 🚀 -- it will be in `3.3.0` 👨‍🚀

PR #700, opened 2021-09-13
feat(settings): tabs scrolls to top if navigated

dVp007:
solving! Scroll to top when switching settings tabs #681

### DIFF
--- a/src/app/components/settings/settings.component.ts
+++ b/src/app/components/settings/settings.component.ts
@@ -46,12 +46,14 @@ export class SettingsComponent implements OnInit {
     private electronService: ElectronService,
     private modalService: ModalService,
     private translate: TranslateService
-  ) {}
+  ) { }
 
   ngOnInit(): void {
     this.additionalInput = this.appState.addtionalExtensions;
   }
-
+  ngOnChanges(): void {
+    document.getElementById('settings').scrollTop = 0;
+  }
   editAdditionalExtensions() {
     this.editAdditional = !this.editAdditional;
     this.additionalInput = this.appState.addtionalExtensions;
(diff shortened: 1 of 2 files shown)


whyboris:
Thank you very much! The code looks like it will work 💯 -- I just want to test it out before merging 🤝

src/app/components/settings/settings.component.ts
```diff
@@ -46,12 +46,14 @@ export class SettingsComponent implements OnInit {
...
     private translate: TranslateService
-  ) {}
+  ) { }
 
   ngOnInit(): void {
     this.additionalInput = this.appState.addtionalExtensions;
   }
-
+  ngOnChanges(): void {
+    document.getElementById('settings').scrollTop = 0;
```
whyboris, line 55: I added `console.log()` above this and can confirm that this method gets called every time you switch between the _Tabs_ in _Settings_, but it doesn't seem to actually scroll to the top 😕 

ps, when you add `ngOnChanges()` method to an Angular class, you should also add it to the class declaration:
```ts
export class SettingsComponent implements OnInit, OnChanges {
```

whyboris (changes requested):
Unfortunately with your changes, the app does not scroll to the top of the tab when switching tabs.

I wonder if this will be useful: https://stackoverflow.com/a/46348211/5017391
`.scrollIntoView()`

I tried using it with your code but it still didn't work. I'm unsure what is going wrong. Please let me know if you'd like help with this 🤝

dVp007:
Okay will check it before can you send me a video or a GIF so that I can figure out what is wrong. Because when I tried it was scrolled at the top.

whyboris:
To replicate:

scroll down to bottom, now click on the 1st tab:

notice the scrollbar is not at the top.
On your branch the scrollbar seems to move up a bit, but not to the top of the tab

dVp007:
Can you please check i actually forgot to commit changes of a file. I guess it should work now.

Issue #682, opened 2021-07-28
JPEG XL (.jxl) support


whyboris:
🎉 JPEG XL is the most-exciting image format in several decades! Let's get VHA to use it 😎 
Expected: 20% size saving on the images VHA generates - and likely faster decoding than JPEG 🚀 

What we're waiting for: `FFmpeg` needs to support `.jxl` output first 😅 

After that we're ready 🏃‍♂️

whyboris:
FFmpeg has merged JPEG XL support but we'll need to wait for the binaries to be available through the `npm` package I'm using ⏳

whyboris:
Fantastic news 🥳  - finally after _years_ of waiting: Chrome Canary (M145) has `jxl` support: [tweet](https://x.com/jonsneyers/status/2011058709817065955)

whyboris:
Does not affect _VHA_ but great news anyway: _Firefox_ `152` now has `jxl` support behind a flag 🥳

PR #702, opened 2021-09-15
JPEG XL - .jxl - works 🎉

whyboris:
*SUPER* exciting 🤩 

This will be waiting for `FFmpeg` to catch up 😅 but I'd like to keep this PR on standby as a reminder of the brighter future 😎 

#682 🚀


whyboris:
Just need the new FFmpeg binary to be available through the `npm` package I'm using ⏳

whyboris:
I intend to get `jxl` into _VHA_ when I get _facial recognition_ functionality working -- that's when `.vha2` file format will need to be updated to include more data, and it will be easier to include a migration script when going from VHA `3` -> `4` 🥳

Issue #655, opened 2021-05-20
Hidden bug in ./node/main-support.ts function hashFileAsync()


geppi:
There is a bug sleeping in the code that calculates the file hashes.
The current version of nodejs used by VHA/Electron hides it but the current version 16.2.0 of nodejs would reveal it.
The problematic line of code is 400 in ./node/main-support.ts:

`fs.read(fd, data, sampleSize, sampleSize, fileSize / 2, (err, bytesRead, buffer) => {`

which intends to load a _sampleSize_ chunk of data from the middle of the file.
This works for even _fileSize_ numbers but for odd numbers the division does return a float while _fs.read_ does expect an integer in its position parameter.

The way this is handled differs depending on the version of nodejs.
In version 12.14.1 and up to version 15.6.0 the code in fs.js simply replaces the float with -1, which results in reading the data from the current file position. Up from version 15.7.0 an exception is thrown if a float is passed in the position parameter.

Unfortunately in the current version of VHA reading the data from the current file position means that the first _sampleSize_ chunk is read again. So while for files with even _fileSize_ the data used in the hash calculation is:

|First _sampleSize_ bytes| + |Middle _sampleSize_ bytes| + |Last _sampleSize_ bytes| + |_fileSize_ number|

for files with odd fileSize the calculation uses:

|First _sampleSize_ bytes| + |First _sampleSize_ bytes| + |Last _sampleSize_ bytes| + |_fileSize_ number|
 
The question is how to fix this ?

The easy way would be to manifest the current behavior explicitly in the code, i.e. check if _fileSize_ is odd and explicitly duplicate the _sampleSize_ data from the beginning of the file. This will keep the hash values identical to their current values.

However, to me it feels like carving an initial flaw into stone. It will for sure increase the probability that the hashes of two different, odd _fileSize_ clips would collide, although this probability is admittedly still very small. The files would for example have to have the same _fileSize_ and pure black frames at the beginning and at the end with identical metadata. Nevertheless this implementation would for sure not be what was intended but the least painful.

The proper fix in my mind would be to convert the resulting float down to the next integer and use a _sampleSize_ chunk from the middle of the file also for odd _fileSize_ clips. However, this would lead to different hashes for those files compared to the current version. Not a problem when deploying a new VHA but definitely an issue if you want to upgrade an existing one.

Maybe your work of updating VHA to Electron 12 would be a good opportunity to fix this although the nodejs version 14.16.0 that it will use does still not reveal the bug.

[...]

Issue #643, opened 2021-02-16
Error in hash function


RedAero:
TL;DR up front: the part of the hash function that is meant to read the middle of the file uses `fileSize / 2` to seek to the middle, however that returns a fraction for odd fileSize values, which is not a valid position for fs.read(), which in return sets the value to zero, ipso facto the beginning of the file is read twice. Apparently passing fractional positions to fs.read() will throw an error in up-to-date versions of node.

https://github.com/whyboris/Video-Hub-App/blob/45f91c7f37c1ddf984e7ff329ae58194fabdb70c/node/main-support.ts#L400

I have been trying to reproduce the hash function in Python so that I can insert files into the JSON DB manually and transfer tags to compressed versions of old files, and after hours of attempts I was at my wits end. It seemed, though I wasn't sure, that my function returned the correct hash for files of even length, but not for odd ones. Obviously my interpretation of the above line was at fault, but I just couldn't figure out **how**.  Long story short, after having exhausted all off-by-one variations, I consulted the wise men of the Freenode #javascript channel, where someone eventually realized that, depending on the version of node used (!), said function may throw an error (apparently in the latest version) or it may just set the position value to -1 (?!). This is (as best I can tell), not mentioned in the [docs](https://nodejs.org/api/fs.html#fs_fs_read_fd_buffer_offset_length_position_callback), nor does it make any sense to me - if the function expects an int, how is -1 going to help? Anyway, as it turns out, what it actually does is set the position to **0**, to the beginning of the file, and modifying my Python code accordingly *did* return the expected hash values. Clearly, this is not intended behaviour - `fileSize / 2` should instead be rounded down (alá Python's // operator).

Also, while I'm here I may as well also mention: there's a lot of needless back-and-forth conversion of values going on in that function.. I'm not familiar with JS, but I would assume it's possible to convert an integer to a byte-like Buffer object directly instead of going through string, and likewise, I would be surprised if the md5 function itself couldn't take byte-like objects. To replicate the results in Python I had to follow the same steps which, at least in Python, is highly redundant. I could have convert the file size (an int) to bytes in one step, concatenated it with the rest , and fed it straight to the md5 function.

Finally, in case anyone would like to follow in my footsteps, here is the Python code for the hash function:

```
def md5_hash(path):
    sample_size = 16 * 1024
    sample_threshold = 128 * 1024
    file_size = os.path.getsize(path)
    with open(path, "rb") as f:
        if file_size < sample_threshold:
[...]

PR #708, opened 2021-10-20
Version bump February 2022

whyboris:
Bumping to _Electron 17_ 🚀 

⚠️ _merge warning_ ⚠️ Fixing #643 and #655 😅 -- though more work needs to be done since now some hashes (about half :trollface: ) will not match old ones 😓 

Let's also use `strictTemplates` #715 👍  and update to _Angular 13_ 🚀 

Let's get `FFmpeg` version `5` into this release too 🎉  #737


whyboris:
🎉 `FFmpeg` version 5 is out when we upgrade `ffmpeg-static` to version `5.0.0` 🚀 

https://github.com/eugeneware/ffmpeg-static/releases/tag/b5.0

Issue #712, opened 2021-11-27
Add kb/s for videos in details


whyboris:
Seems easy to add, may be useful to some 🤔

jozefkopcak:
Hello, could I be assigned to this issue? :)

jozefkopcak:
Hello Boris, I'm a bit lost in the project. I have inserted "bitrate" to ImageElement interface and have made function getBitrate in main-support.ts. I am struggling with what to do next, I can't find where exactly to use it now to make it show in video details. Could you help me? :)

whyboris:
You're doing great! I'm glad you were able to get the bitrate from the `FFprobe` output 🎉 

The last part should be rather straight-forward 🤞 
Once the bitrate is part of the ImageElement - you just have to display it:
https://github.com/whyboris/Video-Hub-App/blob/main/src/app/components/meta/meta.component.html#L37
Add a line there and it should show up for every _details_ view 😎

jozefkopcak:
![image](https://user-images.githubusercontent.com/63211669/145851647-b220f182-b6f3-4940-93f9-c370aa86a68e.png)

I think i got it, my formula for calculating bitrate was ((fileSize/1000)/duration). Do u want it like that? Ot should i change it to ((fileSize/1024)/duration)? In the end, I rounded it up, so it looks better.

whyboris:
Looks great 😍 

Rounding up is a great idea. Calculation I'm unsure whether `1000` or `1024` is the right approach - I wonder if Windows does it correctly (when you right-click on video, go to _details_ and see the bitrate). But don't sweat it too much, I could double-check later too 👌 

I wonder if it's most-common nowadays to have bitrates of closer to 5,000 - so maybe `mb/s` rather than `kb/s` is better? 🤔

jozefkopcak:
![image](https://user-images.githubusercontent.com/63211669/145854120-d2358291-4e34-4399-b04d-936f95b779b8.png)

(my windows is in my native language, dont get too scared :D )
So, windows shows in video details, that bitrate is 1205 kbps which converted to kb/s is 150.6, rounded up is 151 kb/s.

![image](https://user-images.githubusercontent.com/63211669/145854334-b62592c0-6e6e-44d8-ac3e-3863b6094936.png)

So that means, my calculations should be okay :D 

If u want mb/s, it's not a problem. I know, when u have like 60GB Blurays, it can be a bit too much numbers :D

jozefkopcak:
[...]

PR #714, opened 2021-12-13
Showing bitrate in video details 

jozefkopcak:
This is what I ended up with, I hope it satisfies :)

### DIFF
--- a/src/app/components/meta/meta.component.html
+++ b/src/app/components/meta/meta.component.html
@@ -33,7 +33,9 @@
     | {{ video.width }} x {{ video.height }}
 
     | {{ video.fileSizeDisplay }}
-
+    
+    | {{ video.bitrate + ' mb/s' }}
+    
     {{ video.fps ? '| ' + video.fps + ' fps' : '' }}
 
     | <input
--- a/interfaces/final-object.interface.ts
+++ b/interfaces/final-object.interface.ts
@@ -30,6 +30,7 @@ export interface ImageElement {
   duration: number;              // number of seconds - duration of film
   fileName: string;              // full file name with extension - for opening the file
   fileSize: number;              // file size in bytes
+  bitrate: number;               // bitrate of the displayed video file - (fileSize/duration)*1024 
   fps: number;                   // base frame rate of the video in fps
   hash: string;                  // used for detecting changed files and as a screenshot identifier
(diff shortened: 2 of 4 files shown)


src/app/components/meta/meta.component.html
```diff
@@ -33,7 +33,9 @@
     | {{ video.width }} x {{ video.height }}
 
     | {{ video.fileSizeDisplay }}
-
+    
+    | {{ video.bitrate + ' mb/s' }}
```
whyboris, line 37: I'm happy to accept as is. But currently if `bitrate` property is absent (as it would be for old Hubs), there would be an empty display. I recommend something like the line below, for example:
`{{ video.bitrate ? '| ' + video.bitrate + ' mb/s' : '' }}`

whyboris (approved):
Looks great 🎉 
Thank you for your contribution 🙇 🤝 😁

src/app/components/meta/meta.component.html
jozefkopcak: There will always be a bitrate property, because it's calculated from videoSize and duration, which are present in every video there is (if I understand correctly).

whyboris: Oh you're right! I'm sorry, in my mind, before I saw your code, I was thinking you were grabbing the bitrate from what `FFmpeg` was returning (when initially scanning the video), and then I commented based on my guess about your code rather than your code 😓 

I like this implementation rather than what I was thinking, since it provides previously-scanned videos with a bitrate estimate!

I'll merge this in soon 👏

whyboris:
Closes issue #712 🙇 🤝 😁 Thank you!

whyboris:
Thank you @jozefkopcak for the lovely contribution to Video Hub App 🙇 

I might tweak it a bit before release (I think the standard way people think about bitrate is using bits, so I just need to [multiply by 8](https://write.corbpie.com/how-to-calculate-video-bitrate-from-video-size-and-length/) in your formula 😅). 

To this day I get frustrated between "megabits" and "megabytes" - and while I like seeing how many megabytes are used per second, I think most video people expect megabits 🤷 (unsure why - perhaps because this measure was easier to think about when streaming over the internet, which for some annoying reason was in bits not in bytes :trollface: )

PR #742, opened 2022-03-18
add manual tags to magic search

litelime:
The reason for this is I found myself often using the magic search with the assumption that it would include my manual tags, and being surprised when videos I had added tags to didn't show up.


whyboris:
Thank you! Great addition 🎉 
I added "magic search" as a way to search across folders and files - before manual tags were added to the app. Now that we have tags, I think it's the right thing to have "magic" include it 🤝 

I'll be doing a batch merge bugfix in the coming month or two - I'll merge in before the next (bugfix) release of the app 🙇

PR #749, opened 2022-03-28
new ffprobe

whyboris:
Let's give this one a try 🤞 

Seems to have fixed the `M1` mac hanging problem after the last few PR merges 😅


whyboris:
⚠️ because `strictTemplates` is `true` in `tsconfig.json` the app does not build - erroring out with about 15 problems like:
> Argument of type 'MouseEvent' is not assignable to parameter of the type 'VideoClickEmit'
😅 

I'm merging but will try to fix in another branch without resorting to disabling the `strictTemplates` 🤞 

This will require combing through the code to confirm the `string` / `number` debacle is fixed 😅

PR #752, opened 2022-04-05
Short paths

whyboris:
Works for *VS Code* but not when you `npm start` :trollface:


whyboris:
Tested again today - no success, same problems as before [see writeup](https://github.com/maximegris/angular-electron/issues/649#issuecomment-1089141896) :trollface: 

🤷‍♂️  I think the _service worker_ causes some problems I wouldn't otherwise have 🤔

whyboris:
Tested again today (Angular updated several times since before), still doesn't work :trollface:

whyboris:
Link for research: https://github.com/angular/angular-cli/issues/1465#issuecomment-236054952

Looks like you need `baseUrl` too.
Probably need to add this to all the `tsconfig.---.json` files in the repository. I tried many variations but nothing works still 🤦

whyboris:
🤷  frustrating -- still doesn't work. I'll keep the _PR_ open in case someone figures something out 🤞 

I'll re-visit this in 2027 (or earlier if someone comes up with a solution ❤️ ) 😄

PR #754, opened 2022-04-23
Adding support for images, any maybe other file types? 😄 

cal2195:
Just a WIP, adding support for images.

Should be super simple, just work out how to deal with image strips and clips (basically don't), and allow for toggling video and images individually! 💡 

Thoughts?

### DIFF
--- a/interfaces/final-object.interface.ts
+++ b/interfaces/final-object.interface.ts
@@ -5,6 +5,8 @@ export type AllowedScreenshotHeight = 144 | 216 | 288 | 360 | 432 | 504;
 
 export type ResolutionString = '' | 'SD' | '720' | '720+' | '1080' | '1080+' | '4K' | '4K+';
 
+export type FileType = 'video' | 'image';
+
 export interface SourceFolder {
   path: string;
   watch: boolean;
@@ -25,12 +27,13 @@ export interface FinalObject {
 }
 
 export interface ImageElement {
+  type: FileType;                // type of file
   birthtime: number;             // file creation time
   cleanName: string;             // file name cleaned of dots, underscores,and file extension; for searching. Can be *FOLDER* sometimes
   duration: number;              // number of seconds - duration of film
   fileName: string;              // full file name with extension - for opening the file
   fileSize: number;              // file size in bytes
-  bitrate: number;               // bitrate of the displayed video file - (fileSize/duration)*1024 
+  bitrate: number;               // bitrate of the displayed video file - (fileSize/duration)*1024
   fps: number;                   // base frame rate of the video in fps
   hash: string;                  // used for detecting changed files and as a screenshot identifier
   height: AllowedScreenshotHeight; // height of the video (px)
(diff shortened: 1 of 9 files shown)


whyboris:
Hey Cal! Great to "see" you again 😁 

Thank you for this suggestion / start of a discussion. 

I started (50% finished?) *Simple Image Browser* application: https://github.com/whyboris/Simple-Image-Browser (SIB) which might be just what people keep asking me about when they say they want images in VHA. Before I add image types to VHA, I'd like to get the dedicated image browser released first 😅 .

I've felt rather strongly against mixing image and video browsing inside one application, but I see how someone might have images that are associated with some videos and it may be useful to see those all in one place. Coding it up and making the UI for it feels like would be quite nightmarish for my app. I guess I could create an image-only gallery view which would just show images inside a folder without keeping to the 16:9 video resolution; perhaps, I could just use the functionality of the _Simple Image Browser_ of not even extracting screenshots, but simply displaying images live 🤔 ... but at that point I'd just be making a crappier version of SIB 🤷 

I'll definitely consider it once I've released SIB 🤝

cal2195:
The way I'm thinking of it, is that images are just like a video, but with a single frame. When you think of it that way, there's very little to change, and you get all the benefits of what VHA already does extremely well. 💡 

The code should basically be identical, just with less features available for images. 🤔 

And I'm planning on having a toggle for importing/displaying images or videos, so it's totally optional anyway.

My current plan for phase one anyway, is just that. Not even image viewing in app - just like videos, click an image, and your default image viewer opens. (This already just works actually 😂 ).

Thoughts on this view?

cal2195:
I've a few more features and improvements I'd like to add, so if you'd still like to hold off, I'll create and integration branch on my repo for this and future changes, and still try to make pull requests on your main branch. 😄

cal2195:
> I could just use the functionality of the Simple Image Browser of not even extracting screenshots, but simply displaying images live

Thumbnails are _much_ more performant, especially when images get large (raw files), and if you access them over a network on a NAS somewhere. And it's super easy to piggy back off ffmpeg for this - I see it's already done in VHA for replacing your own thumbnail! 😄

whyboris:
I'm certainly open to this possibility. I guess in the `home.component.html` where we feed all the `imageElements` into the view, the _first_ pipe could be the image / video toggle 🤔 

https://github.com/whyboris/Video-Hub-App/blob/main/src/app/components/home.component.html#L691

And then after that, just about nothing else needs to change 👍 

So you are probably right -- if the weird 16:9 view is acceptable (with black letterboxing; which in _Dark Mode_ would be invisible anyway) then the number of changes we need to get this working might be pretty small 🤔 

Thanks as usual for excellent ideas and _actual additions_ to the app 🙇 

I've not thought through all the implications of having images, but it seems like we'll automatically inherit _tags_ 🤷 which I guess some people want (🤷 ?). 

The one minor worry is about `.vha2` growing mighty large, but I think even with 100,000 items the app should survive fine 🤞 

I'm open to a PR that does something like this -- and/or a variation -- I'm open to recommendations 🤝

whyboris:
Thank you @cal2195 for all the contributions you've made to _Video Hub App_ 🙇‍♂️ 

I hope you're doing great 🌞 

I'm closing this PR as I don't intend to allow images into _VHA_ 😅 

I hope to find time in 2027 to resume work on and release `Simple Image Browser` https://github.com/whyboris/Simple-Image-Browser 🚀

Issue #755, opened 2022-04-27
Consider updating the range sliders


whyboris:
They are a bit janky, so maybe we should use a library 🤔 

https://github.com/angular-slider/ngx-slider for example 🤷

KMadhu479:
HI @whyboris, I want to work on this item, can you assign it to me?

whyboris:
Thank you - please reach out if you have any questions. I’ll be back online on Monday. This is a tricky issue - I’m unsure if it will be easy or possible to use a library 😅 but I hope it will be possible. 

Part of the issue is I’m particular about how the result will look and might resist a change even if implemented - I don’t want to hurt anyone’s feelings - if you work on this issue, please consider reaching out with a screenshot before you spend too much time on it 🙇‍♂️

whyboris:
Would make #740 simpler 🤔

whyboris:
Would fix this bug: https://github.com/whyboris/Video-Hub-App/issues/861

whyboris:
The range selectors are janky and worth updating 😅 -- will see if easy-enough before the next release 🤞

whyboris:
I cleaned up and fixed up the currently-existing sliders 🥳 with #953 😁 

I think I'll stick to my custom-built sliders for the upcoming VHA release, but I'll keep the issue open to revisit in 2027 👍

PR #799, opened 2023-02-26
last-time-played sort order

whyboris:
🚀 easy feature to add - was requested by a customer.

Good idea for the future: slider selection (in case someone wants to look back for a video they know they played 1 week ago) 🤷 -- but might be complicated because of the slider selection. May be worth upgrading to custom sliders first if that will make coding easier: #755


whyboris:
😓 this code didn't account for any hub created prior to the release of this version (sorting by `lastPlayed` didn't work on old hubs with this code change)

I finally fixed it with this PR: #947

PR #805, opened 2023-03-15
Manually add ffmpeg and ffprobe

whyboris:
😓  worse in some ways, but allows for manually choosing which version of FFmpeg to include 🤔 

Currently _Mac_ builds are broken with this -- because Mac complains about the security of executing a file that's already part of the build package :trollface: 

And build sizes are double of what they were before 😓 -- so probably not worth merging this in (unless I figure out how to bundle things better) 😓


whyboris:
This may solve #933 🚀  

If available, I can include _FFmpeg_ that has hardware acceleration - creating a separate release just for _nvidia_ cards 🤔

whyboris:
Ran a bunch of tests, looks like `-hwaccel cuda` doesn't speed up, but instead _slows down_ the extraction process 🤦‍♂️ 

So for now, I intend to use regular _FFmpeg_ without _cuda_ support 👌 

This _PR_ can still merge - as it may be nice to allow manual selection of _FFmpeg_ without having to rely on an auto-downloader 🤔

whyboris:
If I find time I will try to build `FFmpeg` from source with the `--disable-encoders` flag -- since VHA only needs to _decode_ (will need to check if this stripped down FFmpeg can still write a `jpg`).

If it works out, I can have VHA's build be much smaller 🚀 

Currently eager to release `3.3.0` so this will have to wait for another release 🙆‍♂️

whyboris:
Couldn't build version `3.3.0` on a _Mac_ for a while :trollface: 

Turned out I could use `process.resourcesPath` string that gets added via _Electron_

Using this branch: `mac-ffmpeg` 🚀  see the _PR_ #990

Issue #871, opened 2024-10-31
[FEATURE] speed up auto-tag generation


whyboris:
Looks like JS `for` loop is far faster than `forEach` (see [video](https://www.youtube.com/watch?v=_pWA4rbzvIg))

The slowest part of the auto-tag generation process is already outsourced to a _web worker_ but it's currently using `forEach`

- https://github.com/whyboris/Video-Hub-App/blob/main/src/app/components/tags-auto/tags.worker.ts#L31
- https://github.com/whyboris/Video-Hub-App/blob/main/src/app/components/tags-auto/tags.worker.ts#L59
- https://github.com/whyboris/Video-Hub-App/blob/main/src/app/components/tags-auto/tags.worker.ts#L89

We may be able to get 2x the speed with a simple `for` loop 🤞 🚀 🤞

riddhi-b:
Hi! can i claim this issue?

whyboris:
Hey @riddhi-b - you've claimed the issue ✅ 🤝 

If you can, could you also perform some sort of a speed test to see if the change is worth it? 🙇‍♂️

riddhi-b:
@whyboris Thanks! Sure I will try, is there a specific tool or anything you like to use to perform speed checks or are you ok with me using anything I can find?

whyboris:
@riddhi-b - I suspect [`console.time()`](https://developer.mozilla.org/en-US/docs/Web/API/console/time_static) is the easiest, though feel free to use whatever you'd like 👍

whyboris:
Hey @Deepak-SBaghel - I believe this was fixed with this PR: https://github.com/whyboris/Video-Hub-App/pull/875

whyboris:
I ran `console.time()` / `console.timeEnd()` and the performance was the same with this change #875 😔 

looks like going from:

```ts
potentialTwoWordMap.forEach((val: number, key: string) => {
```

to 

```ts
 for (const [key, val] of potentialTwoWordMap.entries()) {
```

does not speed things up 🐌

whyboris:
I'll keep this issue open if anyone has other ideas about speeding up the _auto-tag generation_ 

The related code has numerous nested _for_ loops; perhaps we can cleverly use another _Map_ rather than iterating across _arrays_ 🤷

PR #875, opened 2024-12-08
Speed up Autogeneration

riddhi-b:
Related issue: #871

Working on testing to see if this actually did speed anything up.
Changing forEach loops to for loops should speed up performance according to the information in the issue.


whyboris:
Thank you @riddhi-b for the contribution 🤝 

I ran `console.time()` / `console.timeEnd()` and the performance was about 14 seconds on my hub with the code before your change and still about 14 seconds after your code change 🤔 

I was hoping this would be a quick-and-easy win, but it didn't work out 😔 

I'm closing this PR and the associated _Issue_ as it doesn't look like we can speed things up _this easily_ :trollface: 

Cheers 🌞

Issue #840, opened 2024-01-22
Fix incorrect column count when app zoom level set


whyboris:
The number of videos per gallery row works as desired, but if you use the "Zoom" feature (to increase the size of the UI) it seems that upon starting the app, the number of videos per row is one less.

Play around replicating the problem and then fix 🤞

Nat-FL:
Hello! I am interested in working on this issue if that is okay.

whyboris:
Thank you @Nat-FL for the offer 😁 

Please let me know if you'd like any pointers about where to look in the code 🤝 or any other advice.

Cheers 🥳

whyboris:
I believe I fixed it with https://github.com/whyboris/Video-Hub-App/commit/3576c247f1480205c5deca8316b431812d20b884 -- using the code from #899 👍

PR #899, opened 2025-06-18
Fix: Correct column calculation for zoomed-in displays

Sasiprakash-UoA:
The issue I have addressed: #[840](https://github.com/whyboris/Video-Hub-App/issues/840)

What I Did:

- Restored the saved zoom level using webFrame.setZoomFactor(currentZoomLevel)

- After applying zoom, forced a gallery layout refresh using computePreviewWidth() and virtualScroller.refresh() inside a setTimeout(..., 0)

- This ensures the correct number of columns is rendered even when a zoom level like 150% is used

- The change eliminates the issue of one column being missing on cold start due to delayed zoom application

- Verified proper layout at both 100% and 150% zoom levels

- Confirmed that dynamic resizing continues to work smoothly at all zoom levels

Big cheers and thanks for reviewing this! Hope it keeps the gallery looking sharp across zoom levels.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,6 @@
         "@ffprobe-installer/ffprobe": "2.1.2",
         "@iharbeck/ngx-virtual-scroller": "17.0.2",
         "@ngx-translate/core": "15.0.0",
-        "@tweenjs/tween.js": "^25.0.0",
         "an-qrcode": "1.0.7",
         "async": "3.2.6",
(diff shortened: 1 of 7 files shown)


whyboris:
Thank you for this fix! 

I'm closing the PR since I manually added the code you suggested (see commit: https://github.com/whyboris/Video-Hub-App/commit/3576c247f1480205c5deca8316b431812d20b884 ) 🙇‍♂️

PR #949, opened 2026-06-15
ngIf to @if

whyboris:
🤷  modernizing code one file at a time 🐌


whyboris:
Time to use `ng generate @angular/core:control-flow` 🚀

PR #979, opened 2026-06-27
Continuing fixes

whyboris:
- fix word cloud bug when switching from hub to hub 👍 
  - _the bug:_ open word cloud, click some word, click in "recent hubs" to switch to another hub, the word cloud wouldn't update!
- hack-fix manual tag bug when switching from hub to hub 😅 
  - _the bug:_ open manual tags, click some tag, click in "recent hubs" to switch to another hub, the tags would update but clicking wouldn't update UI; clicking in the search filter on the previously-clicked tag wouldn't remove it 😓

### DIFF
--- a/interfaces/shared-interfaces.ts
+++ b/interfaces/shared-interfaces.ts
@@ -35,13 +35,13 @@ export const AllSupportedBottomTrayViews: SupportedTrayView[] = [
 
 // Mouse click events
 export interface VideoClickEmit {
-  mouseEvent: MouseEvent;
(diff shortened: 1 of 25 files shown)


whyboris:
2 weeks later - code looks good 😅  - spent some time on my **Simplest File Renamer**

https://github.com/whyboris/Simplest-File-Renamer

Migrated from _Electron_ to _Tauri_ and the build size went down from 130mb to 3mb 🚀 

Will consider migrating _Video Hub App_ someday to make a slimmer, faster application 🤔

PR #980, opened 2026-07-11
Add Segments view: multi-timestamp video preview grid

kelvincht:
## Summary

Adds a new gallery view, **Segments**, that shows every extracted clip snippet of a video side by side in one row — all sampled timestamps visible and playable at once — complementing the existing Filmstrip (static, all timestamps) and Clips (single playing preview) views.

- Reuses the existing `clips/<hash>.mp4` file (no new extraction, no `.vha2` schema change): the clip already concatenates N uniform snippets back to back, so each cell just seeks/loops within its own segment window client-side.
- Default mode parks each cell on its still frame (near-zero CPU); hover-to-play and an opt-in "autoplay all" mode reuse the existing Clips-view settings buttons and pause on window blur.
- Clicking a segment opens the video at its real source timestamp.
- Clip encoding now sets `-pix_fmt yuv420p -movflags +faststart` so newly-extracted clips are hardware-decodable and start instantly.
- Raised the wizard's clip-snippet cap from 15 to 20.
- Registered like the other views: ribbon button, per-view zoom, keyboard shortcut (`8`), settings persistence.

## Compatibility

`.vha2` and extracted assets are untouched. New `settings.json` keys are additive with fallbacks for older files, and an unrecognized `currentView` value falls back to Thumbnails instead of showing a blank gallery. Verified with real pre-feature and post-feature builds round-tripping the same `settings.json` in both directions with no crashes or data loss.

## Test plan

- [x] `tsc --noEmit` (renderer + main process)
- [x] Production Angular build
- [x] Manual verification in a packaged build: import → Segments view renders correct per-snippet timestamps, hover/autoplay loop within each snippet's window, zoom, single-view-toggle, click-to-timestamp
- [x] Forward/backward settings compatibility verified against a real build of the pre-feature codebase (round-tripped a real `settings.json` old → new → old)


whyboris:
@kelvincht OH WOW 🤩 I'll try this out - sounds interesting from the description 😁 

I suspect I'll like it enough to merge in before the `3.3.0` release 🚀 🤞 

I just finished updating my https://github.com/whyboris/Simplest-File-Renamer and am trying to get https://github.com/whyboris/Simple-Image-Browser released at the same time as _Video Hub App_ - probably mid August (all three at the same time). I'll likely get to this PR later this month / early next month.

Please let me know if you prefer I try before then 🤝 

Cheers ❤️

PR #981, opened 2026-07-12
[stacked on #980] Reduce autoplay CPU/heat + release off-screen media

kelvincht:
## Stacked on #980

This branch continues on top of #980 (Segments view), since it modifies `segments.component.ts` directly. Until #980 merges, this diff will show both PRs' commits combined — please review/merge #980 first, or just look at the last 3 commits here (`74cff8c`, `850e4d9`, `7be8959`) if reviewing standalone.

## What this addresses

Sustained heat/stuttering after a few minutes of gallery autoplay on machines with a finite hardware video-decode pool (e.g. Intel Macs with a discrete GPU), without falling back to static images — video should keep playing as video, just via a more efficient path.

## Changes

- **`VideoAutoplaySchedulerService`** (new): defers each autoplay video's `.play()` call to `requestIdleCallback` (with a ~1.5s timeout + small jitter) instead of a hard concurrency cap. A fast scroll-through never pays for decode on rows already scrolled past (the pending start is cancelled before it fires), but nothing is ever permanently denied — once idle, everything visible gets its turn.
- **`clip.component`** brought up to the optimization level already present in `segments.component`: `preload="metadata"` (was defaulting to eager `preload="auto"` for every visible clip), staggered start for the folder-preview autoplay block (previously only single-clip autoplay staggered), and an explicit `ngOnDestroy` that releases decoders deterministically.
- **`fetchpriority="low"`** on automatic (non-hover) autoplay video reads — libraries on a network mount (SMB, NAS) make I/O contention a real factor here, not just CPU/GPU. Hover-triggered playback is unaffected (still prioritized normally, since it's a direct user action).
- **`IntersectionObserver`-based release/restore** across all six gallery views (Thumbnail, Clip, Filmstrip, Full, Details via `<app-thumbnail>`, Segments): a row that's still mounted but scrolled fully out of the visible viewport — ahead of `virtual-scroller`'s own `[bufferAmount]="0"` culling actually catching up — gets its video paused/unloaded or image `src` cleared immediately, and restores automatically the instant it's visible again. (Caught and fixed a real bug here during testing: `<virtual-scroller>` scrolls internally rather than scrolling the page, so the observers must use it as `root` — the default `root: null` observes the browser's top-level viewport and would never fire.)

## Verified

- `tsc --noEmit` and full `ng build` both clean
- Packaged build, 20-video test hub: with Clips-view autoplay on, confirmed no artificial cap (all visible clips end up playing); confirmed rows fully scrolled off-screen show no `src` and are paused, restoring and resuming on scroll-back; confirmed hover always plays immediately regardless of scheduling state
- Investigated (and did **not** ship) a macOS `h264_videotoolbox` hardware-encode path for clip generation — empirically, VideoToolbox's encoder fails on the `concat`+`scale` filter graph every clip in this app requires, even via the `scale_vt`/`hwupload` hardware-native route, so it would fail on every extraction before falling back. Not worth the wasted overhead; this only would have affected extraction-time heat anyway, not playback.


kelvincht:
Closing per author's preference -- keeping this optimization in my fork only for now, not proposing it upstream.

PR #991, opened 2026-09-16
perf(tags-auto): rewrite two-word tag generation as single-pass Map-based algorithm

whju8812:
## What

The slow part isn't the forEach vs for loop (#875 already tried that and it didn't help) - it's that `getPotentialTwoWordTags` and
`getCleanTwoWordMap` were re-scanning the whole file list once per word / per candidate pair. I rewrote both to only go through the files once, using Sets/Maps instead of repeated `.includes()` calls.

Benchmarked on a synthetic 10k-30k file dataset: ~9x faster for `getPotentialTwoWordTags`, ~45-100x faster for `getCleanTwoWordMap`, same output as before on normal vocabularies.

Related issue: #871


whyboris:
Thank you so much! 

I'm traveling for another week or two but will try to look into this shortly after I'm back 😁 🤝 😄

